PR #8, opened 2019-12-17
Add Geocoding APIs

jsani-radar:
https://app.clubhouse.io/radarlabs/story/989/add-geocoding-apis-to-web-sdk

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -163,6 +163,109 @@ class Radar {
       }
     );
   }
+
+  static geocode(query, callback) {
+    const publishableKey = Cookie.getCookie(Cookie.PUBLISHABLE_KEY);
+
+    if (!publishableKey) {
+      if (callback) callback(STATUS.ERROR_PUBLISHABLE_KEY);
+
+      return;
+    }
+
+    const qs = `query=${query}`;
+
+    const host = Cookie.getCookie(Cookie.HOST) || DEFAULT_HOST;
+    const url = `${host}/v1/geocode/forward?${qs}`;
+    const method = 'GET';
+    const headers = {
+      Authorization: publishableKey
+    };
+
+    const onSuccess = (response) => {
+      try {
+        response = JSON.parse(response);
+
+        if (callback) callback(STATUS.SUCCESS, response, response.addresses);
+      } catch (e) {
+        if (callback) callback(STATUS.ERROR_SERVER);
+      }
+    };
+
+    const onError = (error) => {
+      if (callback) callback(error);
+    };
+
+    Http.request(method, url, {}, headers, onSuccess, onError);
+  }
+
+  static reverseGeocode(latitude, longitude, callback) {
+    const publishableKey = Cookie.getCookie(Cookie.PUBLISHABLE_KEY);
+
+    if (!publishableKey) {
+      if (callback) callback(STATUS.ERROR_PUBLISHABLE_KEY);
+
+      return;
+    }
+
+    const qs = `latitude=${latitude}&longitude=${longitude}`;
+
+    const host = Cookie.getCookie(Cookie.HOST) || DEFAULT_HOST;
+    const url = `${host}/v1/geocode/reverse?${qs}`;
+    const method = 'GET';
+    const headers = {
+      Authorization: publishableKey
+    };
+
+    const onSuccess = (response) => {
+      try {
+        response = JSON.parse(response);
+
+        if (callback) callback(STATUS.SUCCESS, response, response.addresses);
+      } catch (e) {
+        if (callback) callback(STATUS.ERROR_SERVER);
+      }
+    };
+
+    const onError = (error) => {
+      if (callback) callback(error);
+    };
+
+    Http.request(method, url, {}, headers, onSuccess, onError);
+  }
+
+  static ipGeocode(callback) {
+    const publishableKey = Cookie.getCookie(Cookie.PUBLISHABLE_KEY);
+
+    if (!publishableKey) {
+      if (callback) callback(STATUS.ERROR_PUBLISHABLE_KEY);
+
+      return;
+    }
+
+    const host = Cookie.getCookie(Cookie.HOST) || DEFAULT_HOST;
+    const url = `${host}/v1/geocode/ip`;
+    const method = 'GET';
+    const headers = {
+      Authorization: publishableKey
+    }
+
+    const onSuccess = (response) => {
+      try {
+        response = JSON.parse(response);
+
+        if (callback) callback(STATUS.SUCCESS, response, response.regions);
+      } catch (e) {
+        if (callback) callback(STATUS.ERROR_SERVER);
+      }
+    }
+
+    const onError = (error) => {
+      if (callback) callback(error);
+    };
+
+    Http.request(method, url, {}, headers, onSuccess, onError);
+  }
 }
 
 export default Radar;


src/index.js
```diff
@@ -163,6 +163,109 @@ class Radar {
...
+
+      return;
+    }
+
+    const qs = `query=${query}`;
+
+    const host = Cookie.getCookie(Cookie.HOST) || DEFAULT_HOST;
+    const url = `${host}/v1/geocode/forward?${qs}`;
+    const method = 'GET';
+    const headers = {
```
jsani-radar: these are the current `headers` for the `trackOnce` function as well -- do we need to update these to match our other SDKs? We'd need to add `Content-Type`, `User-Agent`, and `X-Radar-Config`

src/index.js
```diff
@@ -105,7 +105,9 @@ class Radar {
 
         // Setup http
         const headers = {
-          Authorization: publishableKey
+          Authorization: publishableKey,
+          'X-Radar-SDK-Version': SDK_VERSION,
+          'X-Radar-Device-Type': 'Web',
```
kochis: Let's set the default headers in `http.js`. We can merge them with whatever is passed in.

src/index.js
```diff
@@ -163,6 +165,203 @@ class Radar {
...
+    const publishableKey = Cookie.getCookie(Cookie.PUBLISHABLE_KEY);
+
+    if (!publishableKey) {
+      if (callback) callback(STATUS.ERROR_PUBLISHABLE_KEY);
+
+      return;
+    }
+
+    const finalLimit = Math.min(limit, 100);
+    let qs = `latitude=${latitude}&longitude=${longitude}&radius=${radius}&limit=${finalLimit}`;
```
kochis: Rather than building the query string in each function, let's do this in `http.js`. 

Maybe change the `body` param to `data`, and the params can be passed in as an object (and converted to qs for the correct http methods)

src/index.js
```diff
@@ -163,6 +165,203 @@ class Radar {
       }
     );
   }
+
+  static searchPlaces(latitude, longitude, radius, chains, categories, groups, limit, callback) {
+    const publishableKey = Cookie.getCookie(Cookie.PUBLISHABLE_KEY);
+
+    if (!publishableKey) {
+      if (callback) callback(STATUS.ERROR_PUBLISHABLE_KEY);
```
kochis: nit: let's keep the braces on if statements for consistency. We should probably add some basic linting at some point.

src/index.js
jsani-radar: hmm, not sure I agree with this 100% -- there's custom logic around certain parameters for the `qs`. For example, capping the limit at 100, turning arrays into strings, etc. I don't want to bake these rules into the request data transformation layer if they won't apply for every request

jsani-radar: yeah, it's a bit odd / clunky at the moment that for all the `GET` endpoints, I have to send an empty body `{}`, and we're also conflating passing data with passing the raw endpoint URL ✔️

src/index.js
jsani-radar: captured offline discussion / future direction in this clubhouse ticket -- https://app.clubhouse.io/radarlabs/story/1018

src/index.js
```diff
@@ -76,92 +79,295 @@ class Radar {
       return;
     }
 
-    if (!navigator || !navigator.geolocation) {
+    Navigator.getCurrentPosition((status, { accuracy, latitude, longitude }) => {
+      if (status != STATUS.SUCCESS) {
```
kochis: nit: `!==`

src/index.js
```diff
@@ -76,92 +79,295 @@ class Radar {
...
+        callback(status);
```
kochis: We should probably return here.

src/index.js
```diff
@@ -76,92 +79,295 @@ class Radar {
...
+
+      const host = Cookie.getCookie(Cookie.HOST) || DEFAULT_HOST;
+      const url = `${host}/v1/users/${_id}`;
+      const method = 'PUT';
+
+      const onSuccess = (response) => {
+        try {
+          response = JSON.parse(response);
+          if (callback) {
+            callback(STATUS.SUCCESS, position.coords, response.user, response.events);
```
kochis: I don't think `position` exists here.

jsani-radar: good catch, changing to `{ accuracy, latitude, longitude }`

test/http.test.js
```diff
@@ -1,11 +1,12 @@
 const { expect } = require('chai');
 const sinon = require('sinon');
 
+import SDK_VESRION from '../src/version';
```
nickpatrick: Typo: `SDK_VERSION`

nickpatrick:
@jsani-radar @kochis Want to release this as `1.2.0-beta.1`? I'll do some testing and then we can release `1.2.0` next week.

src/index.js
```diff
@@ -76,92 +79,585 @@ class Radar {
...
+        if (callback) {
+          callback(status);
+        }
+        return;
+      }
+
+      // Get user data
+      const deviceId = Device.getId();
+      const userId = Cookie.getCookie(Cookie.USER_ID);
+      const placesProvider = Cookie.getCookie(Cookie.PLACES_PROVIDER);
```
nickpatrick: `placesProvider` can be removed everywhere

nickpatrick:
For the new functions, should we accept params as separate vars, or just accept a params object? My hunch is that the latter is easier and more flexible.

e.g.,

`Radar.searchPlaces({ chains: 'starbucks', radius: 1000, limit: 10 }, callback)` instead of `Radar.searchPlaces(1000, 'starbucks', null, null, 10, callback)`

kochis:
Yeah, I think objects are the way to go

jsani-radar:
yeah these are all good calls, especially the params object - will get rid of the awkward intermediate params that customers don't want to provide

Will work on getting these changes through so we can get this approved and into an intermediate beta branch

For now, the base branch is `v3`, should we change that to something more aligned with the Web SDK versions, like `release-1.2.0-beta.1`? Or, we can keep the versions aligned with the mobile SDKs. I'm fine with either

kochis:
Let's keep the beta tag on for now.

src/index.js
```diff
@@ -76,92 +72,686 @@ class Radar {
...
+      );
+    });
+  }
+
+  static getDistanceFromLocation(
+    {
+      originLat,
+      originLng,
+      destinationLat,
+      destinationLng,
```
jsani-radar: we probably want to make this cleaner - should we have clients send `origin` and `destination` as `lat,lng` pairs. We would then not need the manual construction on L721-L722

jsani-radar:
@kochis @nickpatrick no rush, but this is good for another pass - let me know if there's any desired modifications, however small, before we release this for GA

kochis:
I think this looks good for a beta release. Now that we have a bunch of new endpoints, we should probably do a bit of refactor/cleanup and get the test coverage up a bit before cutting a major version.

src/index.js
kochis: Yeah, I think so. Let's use the same format as `near`.

jsani-radar:
in all honesty, most of the drop in test coverage is from my struggle to mock `navigator.geolocation.getCurrentPosition` - every function with signature `*WithLocation` takes an explicit lat / lng should have 100% (pretty much). It's the other functions that grab a location fix from the navigator that are problematic

kochis:
Yeah, lets update the origin / destination params.

jsani-radar:
@kochis @nickpatrick any last changes here before merging the beta?

nickpatrick:
@jsani-radar My preference would be to make the function names and param structure identical to React Native. I can jot that down in a doc if helpful.

src/geocoding.js
```diff
@@ -0,0 +1,134 @@
+import * as Cookie from './cookie';
+import * as Http from './http';
+import Navigator from './navigator';
+
+// consts
+import STATUS from './status_codes';
+
+const DEFAULT_HOST = 'https://api.radar.io';
+
+class Geocoding {
```
kochis, line 10: I like this pattern. Lets create a `/api` directory for each of the classes to go into.

jsani-radar: sure thing, was thinking about something like `/services` - `/api` works well too

jsani-radar:
hey @nickpatrick @kochis - spent the better part of the day refactoring this PR. All the logic was originally happening in `index.js`, which would call `http.js` directly to hit our server. There's an extra `api/` layer in between now, so this should be a good bit cleaner, and `index.js` is relatively compact. Should be good for 👀 again.

That being said, the test coverage has gone down because of these new functions / files - I can spend some time getting that up before the major release. There's some work to clean up the docs and add example code for all the new endpoints.

One decision to make is about overloading functions - turns out JS doesn't really support overloading (https://stackoverflow.com/questions/456177/function-overloading-in-javascript-best-practices), and it seems a little clunky to treat the first parameter in a function differently based on whether there are 2 args passed or 1. This applies for `trackOnce`, `reverseGeocode`, and `getContext` and how to call them to get a location fix. Ideally we wouldn't require `trackOnce({}, callback)`.

src/api/context.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  static getContextForLocation({ latitude, longitude }, callback) {
+    const publishableKey = Cookie.getCookie(Cookie.PUBLISHABLE_KEY);
+
+    if (!publishableKey) {
+      callback(STATUS.ERROR_PUBLISHABLE_KEY);
+
+      return;
+    }
```
jsani-radar, line 33: at some point, I'd like to centralize all the `publishableKey` retrieving and validation in `http.js`

src/api/context.js
```diff
@@ -0,0 +1,64 @@
...
+
+        callback(STATUS.SUCCESS, response.context);
+      } catch (e) {
+        callback(STATUS.ERROR_SERVER);
+      }
+    };
+
+    const onError = (error) => {
+      callback(error);
+    };
```
jsani-radar, line 58: same with the `onSuccess` / `onError` handlers - we can probably pass one callback to `Http.request` and it'll send up the response code accordingly

README.md
```diff
@@ -44,20 +44,6 @@ To get a Radar publishable API key, [sign up for a Radar account](https://radar.
...
-
-To disable Places later, call:
-
-```javascript
-Radar.setPlacesProvider(Radar.PLACES_PROVIDER.NONE);
-```
-
 Learn more about [Places](https://radar.io/documentation/places).
 
 ### Identify user
```
corypisano, line 49: Looking really good - my main thought is just making the new apis visible to users, can we add some of the calls to the readme? Right now it only talks about trackOnce

src/api/context.js
kochis: Might makes sense to bite that off now

src/api/context.js
kochis: Yeah, even getting the host. I feel like the call to request should just be:
```js
Http.request('GET', '/v1/context', queryParams, callback)
```

src/index.js
```diff
@@ -66,102 +59,234 @@ class Radar {
...
   }
 
-  static trackOnce(callback) {
-    const publishableKey = Cookie.getCookie(Cookie.PUBLISHABLE_KEY);
-
-    if (!publishableKey) {
-      if (callback) {
-        callback(STATUS.ERROR_PUBLISHABLE_KEY);
-      }
+  static getLocation(callback) {
```
kochis, line 62: Wonder if it still makes sense to expose this. It's just a wrapper around navigator?

src/index.js
```diff
@@ -66,102 +59,234 @@ class Radar {
...
+    if (!callback) {
```
kochis, line 63: We should probably throw an error if there's no callback. Otherwise it wont be clear why nothing is happening.

README.md
nickpatrick: @corypisano @jsani-radar I'd actually like to start moving a lot of this into Contentful so we can host it on `https://radar.io/documentation` and make changes without pushing to GitHub. I'm happy to own docs.

nickpatrick:
@jsani-radar `trackOnce({}, callback)` feels janky, especially since it will most commonly be called without a location.

How about something like

```
const trackOnce = (arg1, arg2) => {
  if (typeof arg1 === 'function') {
    // location not provided
  } else if (typeof arg2 === 'object') {
    // location provided, validate it
  } else {
    // error
  }
}
```

@kochis ^

src/index.js
jsani-radar: correct

kochis:
@nickpatrick Yeah, that's a fairly common pattern actually.

src/http.js
```diff
@@ -1,33 +1,73 @@
...
-export function request(method, url, body, headers, successCallback, errorCallback) {
+function _handleResponse(response, jsonKey) {
+  if (jsonKey) {
+    const res = JSON.parse(response);
+    return res[jsonKey];
+  }
+  return response;
+}
+
+export function request(method, path, data, jsonKey, callback) {
```
jsani-radar, line 16: added a `jsonKey` parameter to grab the nested data in `_handleResponse` above

src/api/track.js
```diff
@@ -0,0 +1,76 @@
...
+
+    const path = `v1/users/${_id}`;
+    const method = 'PUT';
+
+    const httpCallback = (status, response) => {
+      if (status === STATUS.SUCCESS) {
+        try {
+          response = JSON.parse(response);
+          if (callback) {
+            callback(STATUS.SUCCESS, { latitude, longitude, accuracy }, response.user, response.events);
```
jsani-radar, line 58: this callback has a different format / logic from the rest, hence the custom `httpCallback` - the raw response is handled here as `null` is passed for the `jsonKey`

src/http.js
kochis: Rather than passing the json key, could this just be handled in the callback?

src/api/geocoding.js
```diff
@@ -0,0 +1,68 @@
...
+
+    const path = `v1/geocode/reverse`;
+    const method = 'GET';
+
+    const httpCallback = (status, response) => {
+      callback(status, response);
+      return;
+    };
+
+    Http.request(method, path, queryParams, 'addresses', httpCallback);
```
kochis, line 52: nit: to make these methods even more compact, let's just add the variables here

```js
Http.request('v1/geocode/reverse', 'GET', queryParams, 'addresses', httpCallback)
```

kochis:
This looks great! Thanks for adding all the additional tests. Couple of nits, but I think this is ready for a beta release. I have a couple fast-follow ideas, but it would be good to get in a working version so we can test for regressions.

corypisano:
looking great, nice work! Only concern is what we plan to do with the documentation, will we add to contentful and get it hosted on our docs in the next few days? otherwise I'd vote for adding some examples to the readme

at the moment I wouldn't know how to use the sdk without going in and understanding the code, looking at tests

nickpatrick:
@corypisano @jsani-radar Same thing as React Native, I'm thinking we move the docs to Contentful to make them easier to manage. I can knock that out this week.

corypisano:
Ok cool 👍

jsani-radar:
@nickpatrick @corypisano - happy to discuss how we can make managing the docs easier, only potential issue I see with Contentful is managing cross-platform differences (calling functions in Python is different from JS) - do we want separate Contentful pages for each non-mobile SDK?

nickpatrick:
@jsani-radar Yep, I think we need separate docs pages for each platform which I can add

src/version.js
```diff
@@ -1 +1 @@
-export default '1.1.0';
+export default '1.2.0-beta';
```
kochis, line 1: Lets do `3.0.0-beta`

jsani-radar: do we need an intermediate v3.0.0-beta.1 branch, or can this go directly to master?

kochis: It can go to master

PR #16, opened 2020-04-08
move callback code to async

kochis:
Refactor to use async/await. Polyfill for async & promise to support IE 11.

- [x] Use async/await for handling service calls
- [x] Update babel config for ie 11 support
- [x] return http response as last arg
- [x] deprecate `SUCCESS` status. Callback will be `(err, ...)` to better support promisify
- [x] return api errors
- [x] clean up missing code test coverage
- [x] update readme with new callback and error messages
- [x] return objects in callbacks instead of individual arguments


src/api/context.js
```diff
@@ -1,37 +1,21 @@
...
 
-    Http.request('GET', `v1/context`, queryParams,
-      (status, response) => {
-        callback(status, response ? response.context : null);
-        return;
-      }
-    );
+    const response = await Http.request('GET', `v1/context`, params);
+
+    return response.context;
```
jsani-radar, line 18: something I was thinking about - for the API demo page, do we still want to show the `meta` and top-level key? Right now, https://radar.io/demo/api has both

src/api/search.js
```diff
@@ -1,171 +1,125 @@
...
 
-  static autocompleteNear(
-    {
+
+  static async autocomplete(searchOptions={}) {
+    // NOTE: I'm not sure how caching works with getCurrentPosition,
+    // but this could be problematic here if needs to compute each time.
+    // We could add a cache: true flag to retreive previous value
+    if (!searchOptions.near) {
+      const { latitude, longitude } = await Navigator.getCurrentPosition();
```
jsani-radar, line 98: looks like you can give the navigator a `maximumAge` for freshness of the position - https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition

src/navigator.js
```diff
@@ -2,37 +2,35 @@
 import STATUS from './status_codes';
 
 class Navigator {
-  static getCurrentPosition(callback) {
-    if (!navigator || !navigator.geolocation) {
-      callback(STATUS.ERROR_LOCATION, {});
-      return;
-    }
+  static getCurrentPosition() {
+    return new Promsie((resolve, reject) => {
```
jsani-radar, line 6: nit: typo

src/index.js
```diff
@@ -61,239 +61,167 @@ class Radar {
...
-        callback(status);
-        return;
-      }
-
-      callback(status, { latitude, longitude, accuracy });
-      return;
-    });
+    Navigator.getCurrentPosition()
+      .then((location) => {
+        callback(STATUS.SUCCESS, location);
```
jsani-radar, line 69: if we invoke as `callback(null, location)` and likewise throughout this file, I think our top-level functions can be passed to `util.promisify`

src/api/context.js
kochis: Yeah, that's a good point. Will need to think about that a bit.

package.json
```diff
@@ -25,21 +25,24 @@
   },
   "dependencies": {},
   "devDependencies": {
```
jsani-radar, line 27: quite a few dependabot PRs popped up after I merged `3.0.0-beta`, hopefully those get knocked out as well with the fresh `npm i` - I saw `lodash` got bumped to 4.17.15 in the lockfile

src/api/routing.js
```diff
@@ -1,64 +1,37 @@
...
+    if (destination) {
+      destination = `${destination.latitude},${destination.longitude}`;
+    }
+    if (modes) {
+      modes = modes.join(',');
+    }
 
-  static getDistanceWithOrigin(
-    {
+    const params = {
```
jsani-radar, line 27: seems like we're constructing the `params` here with potentially `undefined` values, only to hydrate them out in `http.js` - alternatively, we could have `params = {}` at the top above `origin` and build up the map within the `if` conditions, that would avoid the extra filtering downstream

fine with either approach

src/device.js
```diff
@@ -1,24 +1,28 @@
...
-  }
-
-  // generate new deviceId
-  const uuid = generateUUID();
-  Cookie.setCookie(Cookie.DEVICE_ID, uuid);
-  return uuid;
-}
-
-function generateUUID() {
+const generateUUID = () => {
```
jsani-radar, line 3: nit: internal function not exported, prefix with `_`?

src/http.js
```diff
@@ -1,65 +1,73 @@
...
-    callback(STATUS.ERROR_PUBLISHABLE_KEY);
-    return;
-  }
+      const xhr = new XMLHttpRequest();
+
+      let url = `${API_HOST.getHost()}/${path}`;
+
+      // remove undefined values
+      let body = {};
+      Object.keys(data).forEach((key) => {
```
jsani-radar, line 18: I guess we get this polyfilled now with `commonjs`?

src/http.js
```diff
@@ -1,65 +1,73 @@
...
 
-  xhr.send(JSON.stringify(body));
+      xhr.setRequestHeader('Authorization', publishableKey);
+      xhr.setRequestHeader('Content-Type', 'application/json');
+      xhr.setRequestHeader('X-Radar-Device-Type', 'Web');
+      xhr.setRequestHeader('X-Radar-SDK-Version', SDK_VERSION);
+
+      xhr.onload = () => {
+        try {
+          resolve(JSON.parse(xhr.response));
```
jsani-radar, line 54: doesn't look like we're checking `xhr.status` anymore? ex. the server could return a 429

test/api/geocoding.test.js
```diff
@@ -5,167 +5,77 @@ const sinonChai = require('sinon-chai');
...
 describe('Geocoding', () => {
   let httpStub;
   let navigatorStub;
 
-  const latitude = 40.7041895;
-  const longitude = -73.9867797;
+  const query = '20 Jay Street';
 
-  const mockQuery = '20 Jay Street';
+  const geocodeResponse = { meta: {}, address: {} };
```
jsani-radar, line 22: nit: forward / reverse return `addresses`, ip returns `address` - fine to leave as is since it's just testing though

src/index.js
```diff
@@ -61,239 +61,177 @@ class Radar {
...
+        const { location, user, events } = response;
+        callback(null, location, user, events, response);
+      })
+      .catch(callback);
   }
 
   static getContext(arg0, arg1) {
     if (!arg0) {
-      throw new Error(STATUS.ERROR_PARAMETERS);
+      throw new Error(ERROR.MISSING_CALLBACK);
```
jsani-radar, line 111: will this potentially be an issue if someone does `util.promisify(getContext)`? then, the call would look like `await Radar.getContext()`, so technically no `arg0`

src/index.js
```diff
@@ -94,10 +94,14 @@ class Radar {
       throw new Error(ERROR.PARAMETERS);
     }
 
-    Track.trackOnce(position)
+    Track.trackOnce(location)
       .then((response) => {
-        const { location, user, events } = response;
-        callback(null, location, user, events, response);
+        callback(null, {
```
jsani-radar, line 99: do we want the raw `response` to be part of the object we send back? might be odd to see from the caller's perspective if someone does `Radar.getContext((res) => console.log(res));` and sees double the information - we could keep the full response as the 3rd arg always, since the 2nd will always be an object

src/index.js
kochis: I'm not sure about the internals, but I suspect the promisify wrapper will be giving this a callback (this code will not work without a callback)

package.json
kochis: I'll do those separately. Want to get something working in to make sure we don't regressions.

src/api/routing.js
kochis: I like the filtering logic just being in one place (https) that applies to all functions calling it.

src/http.js
kochis: I don't think there's a need to check since it's just passing the error through to the caller now so it can inspect the api error (and get a better error message).

src/index.js
kochis: That's true. The other thing we could do is include `meta` on on this object. We're doing anything _yet_ with meta, but it might be useful to surface it (eg. if we want to support pagination)

kochis: On second thought, I wonder if we should just return the response object. It will just be the top-level keys + meta anyways.

src/http.js
jsani-radar: true, we can probably prune the error codes to only include what JS can throw - doesn't look like we match against 401, 429, etc anymore

src/index.js
jsani-radar: am fine with either approach - technically we're not consistent with this everywhere, as we don't destructure the individual components of the context response (geofences, place, regions), so I'm fine keeping it generic to return the entire response for now, and building in conveniences later

README.md
```diff
@@ -73,21 +73,24 @@ The SDK uses the [HTML5 geolocation API](https://developer.mozilla.org/en-US/doc
...
-
-- `Radar.STATUS.SUCCESS`: the request succeeded
-- `Radar.STATUS.ERROR_PUBLISHABLE_KEY`: the SDK was not initialized
-- `Radar.STATUS.ERROR_PERMISSIONS`: the user has not granted location permissions for the website
-- `Radar.STATUS.ERROR_LOCATION`: location services were unavailable, or the location request timed out
-- `Radar.STATUS.ERROR_NETWORK`: the network was unavailable, or the network connection timed out
-- `Radar.STATUS.ERROR_UNAUTHORIZED`: the publishable API key is invalid
-- `Radar.STATUS.ERROR_RATE_LIMIT`: exceeded rate limit of 1 request per second per user or 60 requests per hour per user
-- `Radar.STATUS.ERROR_SERVER`: an internal server error occurred
+`error` will be null for a successfull call. Errors thrown by the Radar SKD include:
```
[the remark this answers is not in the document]
jsani-radar: ^ bump

src/index.js
```diff
@@ -9,15 +9,16 @@ import Track from './api/track';
 
 // consts
 import SDK_VERSION from './version';
+import ERROR from './error_codes';
 import STATUS from './status_codes';
 
 class Radar {
   static get VERSION() {
     return SDK_VERSION;
   }
```
jsani-radar, line 19: do we want to add back the getter for `STATUS` as well for `STATUS.SUCCESS`?

PR #18, opened 2020-04-10
response available on error

kochis:
Made a few adjustments to make the success/error cases more consistent:
- [x] In the case of an error, the second arg will still be present as an empty object (this allows for destructuring in signature)
- [x] In the case of an http error, the response will still be present in the third arg

### DIFF
--- a/demo/test.html
+++ b/demo/test.html
@@ -101,7 +101,7 @@
 
         Radar.initialize(publishableKey);
 
-        Radar.trackOnce(function(err, { status, location, user, events, response }) {
+        Radar.trackOnce(function(err, { status, location, user, events }, response) {
 
           console.log({ err, location, user, events, response });
 
--- a/demo/test.html
+++ b/demo/test.html
@@ -101,8 +101,9 @@
 
         Radar.initialize(publishableKey);
 
-        Radar.trackOnce(function(err, { location, user, events, response }) {
-          console.log(err, location, user, events);
+        Radar.trackOnce(function(err, { status, location, user, events, response }) {
+
+          console.log({ err, location, user, events, response });
 
           $('input,button').prop('disabled', false);
           $('.loader').addClass('d-none');
@@ -114,7 +115,7 @@
             return;
           }
 
-          $('#status').html(formatCode(response.meta));
+          $('#status').html(status);
 
           $('#location').html(formatCode({
             accuracy: location.accuracy,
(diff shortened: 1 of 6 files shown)


demo/test.html
```diff
@@ -101,8 +101,9 @@ <h3>Events</h3>
 
         Radar.initialize(publishableKey);
 
-        Radar.trackOnce(function(err, { location, user, events, response }) {
-          console.log(err, location, user, events);
+        Radar.trackOnce(function(err, { status, location, user, events }, response) {
+
+          console.log({ err, location, user, events, response });
```
jsani-radar, line 106: remove this?

src/index.js
```diff
@@ -11,6 +11,26 @@ import Track from './api/track';
 import SDK_VERSION from './version';
 import STATUS from './status';
 
+const handleError = (callback) => {
+  return (err) => {
+
+    // Radar Error
```
jsani-radar, line 17: nit: can we mention the JS SDK explicitly throwing these? slightly confusing as an http error is still one the Radar backend returns

test/index.test.js
```diff
@@ -182,19 +232,21 @@ describe('Radar', () => {
...
       } catch (e) {
         expect(e.message).to.equal('ERROR_MISSING_CALLBACK');
+        done();
       }
     });
 
-    it('should propagate the err if not successful', () => {
-      navigatorStub.rejects(STATUS.ERROR_LOCATION);
+    it('should propagate the err if not successful', (done) => {
+      navigatorStub.returns(Promise.reject(STATUS.ERROR_LOCATION));
```
jsani-radar, line 245: did `.rejects` not work here? also curious why we need `done` here but not in the `api/` files

demo/test.html
kochis: This is on the demo page, I think it's fine to leave it.

test/index.test.js
kochis: `rejects` implicitly wraps everything in an `Error` object, where we're just expecting them to be strings for now.

We need `done` in these tests since it's using callbacks (the api tests are all returning promises)

src/index.js
kochis: We're still explicitly throwing all of them. The only difference is we need to pass back an object for the `http` errors so we can capture the response, so we handle those separately.

PR #21, opened 2020-04-15
Expose setMetadata

jsani-radar:
https://app.clubhouse.io/radarlabs/story/2241/add-setmetadata-to-js-sdk

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -81,6 +81,19 @@ class Radar {
     Cookie.setCookie(Cookie.DESCRIPTION, description);
   }
 
+  static setMetadata(metadata) {
+    if (!metadata) {
+      Cookie.deleteCookie(Cookie.METADATA);
+      return;
+    }
+
+    metadata = JSON.stringify(metadata).trim();
+    if (metadata.length === 0 || metadata.length > 256) {
+      Cookie.deleteCookie(Cookie.METADATA);
+    }
+    Cookie.setCookie(Cookie.METADATA, metadata);
+  }
+
   static getLocation(callback) {
     if (!callback) {
       throw new Error(STATUS.ERROR_MISSING_CALLBACK);
--- a/src/index.js
+++ b/src/index.js
@@ -87,7 +87,7 @@ class Radar {
       return;
     }
 
-    metadata = JSON.stringify(metadata).trim();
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -79,6 +79,19 @@ class Radar {
...
   }
 
+  static setMetadata(metadata) {
+    if (!metadata) {
+      Cookie.deleteCookie(Cookie.METADATA);
+      return;
+    }
+
+    metadata = JSON.stringify(metadata).trim();
+    if (metadata.length === 0 || metadata.length > 256) {
```
kochis, line 89: Why delete if larger than 256?

jsani-radar: ah, I see - that's the cap on `description` in upsert, `metadata` is capped at 16 keys, changing

kochis: We also don't need to be doing any validations client-side. It should fail on the server (no need to duplicate the logic here)

jsani-radar: also realized it's impossible for `JSON.stringify(metadata)` to have length 0 at this branch with the `if (!metadata)` check above, even stringifying `''` yields a nonzero length
```
> JSON.stringify('');
< """"
```

src/index.js
```diff
@@ -81,6 +81,16 @@ class Radar {
...
   }
 
+  static setMetadata(metadata) {
+    if (!metadata) {
+      Cookie.deleteCookie(Cookie.METADATA);
+      return;
+    }
+
+    metadata = JSON.stringify(metadata);
+    Cookie.setCookie(Cookie.METADATA, metadata);
```
kochis, line 91: Small nit, don't reassign the value:
```js
 Cookie.setCookie(Cookie.METADATA, JSON.stringify(metadata));
```

src/index.js
```diff
@@ -73,19 +73,24 @@ class Radar {
       return;
     }
 
-    description = String(description).trim();
-    if (description.length === 0 || description.length > 256) {
+    const sanitizedDescription = String(description).trim();
+    if (sanitizedDescription.length === 0 || sanitizedDescription.length > 256) {
```
kochis, line 77: I know this is legacy code, but I feel like we should drop the length checks here too. It seems like pretty unexpected behavior to just clear it out.

PR #24, opened 2020-05-06
make near param optional in autocomplete

corypisano:


test/api/search.test.js
```diff
@@ -141,6 +141,19 @@ describe('Search', () => {
...
     });
+    
+    describe('navigator is disabled', () => {
+      it('should have near undefined and return an autocomplete response', () => {
+        httpStub.resolves(autocompleteResponse);
+        
+        const disableNavigator = true;
+        return Search.autocomplete({ query, disableNavigator })
+          .then((response) => {
+            sinon.assert.calledWith(httpStub, 'GET', 'v1/search/autocomplete', { query: 'mock-query', near: undefined, limit: undefined });
```
jsani-radar, line 152: nit: can also make this assertion some form of `expect(httpStub).to.be.calledWith` - I believe we do so with some of the cookie / device tests

corypisano: ah nice yeah thanks ill follow the pattern

corypisano:
yea i think `useNavigator` is probably cleaner

corypisano:
https://app.clubhouse.io/radarlabs/story/2578/make-near-optional-in-web-sdk-autocomplete

src/api/search.js
```diff
@@ -72,21 +72,17 @@ class Search {
...
+    // if near is not provided, server will use geoIP as fallback
 
     let {
       query,
       near,
       limit,
     } = searchOptions;
 
-    near = `${near.latitude},${near.longitude}`;
+    if (near?.latitude && near?.longitude) {
```
kochis, line 83: We we using this operator anywhere else?

kochis: Want to make sure this is getting compiled properly

corypisano: ah good call, we arent using it anywhere in js sdk, but just ran the build and looks ok
```
if (((_near = near) === null || _near === void 0 ? void 0 : _near.latitude) && ((_near2 = near) === null || _near2 === void 0 ? void 0 : _near2.longitude)) {
                  near = "".concat(near.latitude, ",").concat(near.longitude);
                }

                params = {
                  query: query,
                  near: near,
                  limit: limit
                };
```
getting expanded to the null checks

PR #30, opened 2020-06-29
add layers into geocoding params

blackmad:


jsani-radar:
out of curiosity - does this need to be added for `autocomplete` as well?

blackmad:
good catch, I missed it because it's under /search. need another review on this

blackmad:
we're going to need to do a version bump for this, yes?

PR #31, opened 2020-06-30
Fix autocomplete tests, bump version

jsani-radar:
https://app.clubhouse.io/radarlabs/story/3346/web-sdk-fix-autocomplete-tests-bump-version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "radar-sdk-js",
-  "version": "3.0.0-beta.8",
+  "version": "3.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "Web JavaScript SDK for Radar, the location context platform",
   "homepage": "https://radar.io",
   "license": "Apache-2.0",
-  "version": "3.0.0",
+  "version": "3.0.1",
   "main": "dist/index.js",
   "files": [
     "dist",
--- a/src/version.js
+++ b/src/version.js
@@ -1,1 +1,1 @@
-export default '3.0.0';
+export default '3.0.1';
--- a/test/api/search.test.js
+++ b/test/api/search.test.js
@@ -19,10 +19,11 @@ describe('Search', () => {
 
   const radius = 100;
   const chains = ['dunkin', 'sbucks'];
-  const categories = ['coffe-shop'];
+  const categories = ['coffee-shop'];
   const groups = ['airport'];
   const tags = ['geofence-tag'];
   const metadata = {'geofence-metadata-key': 'geofence-metadata-value'};
+  const layers = ['venue', 'address'];
   const limit = 50;
   const query = 'mock-query';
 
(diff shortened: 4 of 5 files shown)


package-lock.json
```diff
@@ -4906,7 +4900,15 @@
       "integrity": "sha1-MAV0OOrGz3+MR2fzhkjWaX11yQM=",
       "dev": true,
       "requires": {
-        "minimist": "0.2.1"
+        "minimist": "0.0.8"
```
kochis, line 4903: I don't think these changes should be in here

jsani-radar:
seems like CI never ran on the dev branch 🤔 
Screen Shot 2020-06-29 at 10 16 39 PM

package-lock.json
kochis: https://github.com/radarlabs/radar-sdk-js/commit/f279dd75f0e9ddd946c4f08aa85a5c10e762c8e4

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "radar-sdk-js",
-  "version": "3.0.0-beta.8",
+  "version": "3.0.1",
```
jsani-radar, line 3: does this actually need to be bumped? or only with a fresh `npm i`? doesn't seem like we ever did it for v3

kochis, line 3: ooh, nice catch. Maybe we should test for this too in `version.test.js`

jsani-radar: agreed, but still kinda feels odd to me to be "manually" bumping this value

kochis: Yeah, I don't think we should be manually editing this file. But the test should ensure that `npm install` is run before committing (I assume that's how this field gets updated).

jsani-radar: right, but unfortunately `npm i` locally resulted in those other package version changes... that's why I ended up having to go manual

kochis: Try nuking `node_modules` first? We also might need to add some of those packages to `package.json` to make sure they're installing the correct versions.

jsani-radar: tried doing so on this branch and on master without success - I added the following as dependencies but `npm i` was hellbent on using the old versions of `minimist` / `just-extend`, even going as far as installing those versions independently as subdependencies:
package.json
```
"just-extend": "^4.0.0",
"minimist": "^0.2.1",
```
package-lock.json diff after:
```
@@ -4906,7 +4906,15 @@
       "integrity": "sha1-MAV0OOrGz3+MR2fzhkjWaX11yQM=",
       "dev": true,
       "requires": {
-        "minimist": "0.2.1"
+        "minimist": "0.0.8"
+      },
+      "dependencies": {
+        "minimist": {
+          "version": "0.0.8",
+          "resolved": "https://registry.npmjs.org/minimist/-/minimist-0.0.8.tgz",
+          "integrity": "sha1-hX/Kv8M5fSYluCKCYuhqp6ARsF0=",
+          "dev": true
+        }
       }
     },
     "mocha": {
@@ -5026,10 +5034,18 @@
       "dev": true,
       "requires": {
         "@sinonjs/formatio": "3.0.0",
-        "just-extend": "^4.0.0",
+        "just-extend": "^3.0.0",
         "lolex": "^2.3.2",
         "path-to-regexp": "^1.7.0",
         "text-encoding": "^0.6.4"
+      },
+      "dependencies": {
+        "just-extend": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/just-extend/-/just-extend-3.0.0.tgz",
+          "integrity": "sha512-Fu3T6pKBuxjWT/p4DkqGHFRsysc8OauWr4ZRTY9dIx07Y9O0RkoR5jcv28aeD1vuAwhm3nLkDurwLXoALp4DpQ==",
+          "dev": true
+        }
       }
     },
```

jsani-radar: thinking about this a bit more - these are all `devDependencies`, not sure the version matters if the packages won't make it to the final build

kochis: Yeah, these things are only for local development. Probably good to merge for now, can figure this out later.

PR #32, opened 2020-07-01
one-step publishing

blackmad:


package.json
```diff
@@ -20,7 +20,8 @@
     "build:cjs": "rollup -c rollup-cjs.config.js",
     "build:iife": "rollup -c rollup-iife.config.js",
     "build:iife-min": "rollup -c rollup-iife-min.config.js",
-    "report-coverage": "./node_modules/.bin/codecov"
+    "report-coverage": "./node_modules/.bin/codecov",
+    "postpublish": "PACKAGE_VERSION=$(cat package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[\",]//g' | tr -d '[[:space:]]') echo VERSION: $PACKAGE_VERSION && git tag $PACKAGE_VERSION && git push --tags &&  for i in dist/*; do aws s3 cp $i s3://js.radar.io/$PACKAGE_VERSION/; done && echo do not forget to update README"
```
jsani-radar, line 24: not sure if the trailing `%` would end up causing problems:
Screen Shot 2020-07-01 at 1 45 52 PM

kochis, line 24:
```
cat package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[\",]//g' | tr -d '[[:space:]]'
```
Can we just use `jq` for this?
```
cat package.json | jq .version
```

kochis:
```
for i in dist/*;
```
I don't know if we want `index.js` in the public folder.

kochis:
```
cp $i s3://js.radar.io/$PACKAGE_VERSION/
```
Should be prefixed with `v`
```
cp $i s3://js.radar.io/v$PACKAGE_VERSION/

**Comment by blackmad** [comment]
changed to vXXX, changed to jq, not uploading index.js

File: package-lock.json [hunk 2]
```diff
@@ -5026,7 +5026,7 @@
       "dev": true,
       "requires": {
         "@sinonjs/formatio": "3.0.0",
-        "just-extend": "^4.0.0",
+        "just-extend": "^3.0.0",
```
jsani-radar, line 5029: I don't think we need any `package-lock.json` updates for an additional `npm` script - @kochis unfortunately I think we're gonna keep running into this 😬

PR #33, opened 2020-07-01
CVE serialize javascript 2.1.2

ashalam:


package-lock.json
```diff
@@ -6885,7 +6885,7 @@
       "requires": {
         "@babel/code-frame": "^7.0.0",
         "jest-worker": "^23.2.0",
-        "serialize-javascript": "^1.5.0",
+        "serialize-javascript": "^2.1.1",
```
tjulien, line 6888: would be good to do an npm install to update package-lock.json too

package-lock.json
```diff
@@ -4906,7 +4900,15 @@
       "integrity": "sha1-MAV0OOrGz3+MR2fzhkjWaX11yQM=",
       "dev": true,
       "requires": {
-        "minimist": "0.2.1"
+        "minimist": "0.0.8"
```
tjulien, line 4903: did we accidentally downgrade `minimist`?

ashalam: no, i think this is due to my past PR not having done `npm ci`. Once this serialize issue is fixed i can go back and fix minimist if it pops up again.

tjulien: This will downgrade minimist to a vulnerable version, we should revert it.

PR #39, opened 2020-08-11
reverting serialize-javascript and add CodeQL

ashalam:
- this revert fixes the uglify error. This will however bring back the gh dependency security alert.
- adding CodeQL for code scanning.
- https://trello.com/c/3GpWnbnn/782-set-up-gh-code-scanning

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6885,7 +6885,7 @@
       "requires": {
         "@babel/code-frame": "^7.0.0",
         "jest-worker": "^23.2.0",
-        "serialize-javascript": "^2.1.2",
+        "serialize-javascript": "^1.5.0",
         "uglify-js": "^3.4.9"
       },
       "dependencies": {
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,52 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: master
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: '*'
+
+jobs:
+  analyse:
+    name: Analyse
+    runs-on: ubuntu-latest
+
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -6885,7 +6885,7 @@
       "requires": {
         "@babel/code-frame": "^7.0.0",
         "jest-worker": "^23.2.0",
-        "serialize-javascript": "^2.1.2",
+        "serialize-javascript": "^1.5.0",
```
jsani-radar, line 6888: Understood how this works with `npm ci`, but I think we might have to pin this in `package.json` - otherwise, a new developer who does a fresh `npm i` will retrieve the latest version

ashalam: just reverting so i can move forward with github advanced security. Once that is done i can pin it. lot of the new ones i have already started moving away from modifying lock and adding them in package as a resolution.

jsani-radar: ok, gotcha - I guess since we won't cut a new SDK version from this release, it should be fine /cc @kochis

jsani-radar:
hold on, random thought - could you add GH advanced security to this PR to see if it triggers instead of merging this to `master`?

ashalam:
added codeql, checks pass with the revert

jsani-radar:
I'm a bit confused, what's the intended ending state of this PR? I thought the revert was so that GH security analysis _would_ catch a security alert - is having the checks pass what we want in this scenario?

ashalam:
we are kicking the tires with codeql. I don't have enough info to share on how this would work. This is just the first commit to see how it does what it does.

tjulien:
Was the original intention of this PR to get this other PR to work:

https://github.com/radarlabs/radar-sdk-js/pull/38

If so we should close this one and do the work in the other PR.

PR #45, opened 2021-05-27
Add Trips, device id

corypisano:
Adds:
- Trips! (`startTrip`, `updateTrip`, `completeTrip`, `cancelTrip`, and adds tripOptions to trackOnce)
- set custom headers
- set device id
- option to set custom track endpoint

Story details: https://app.clubhouse.io/radarlabs/story/6419


src/api/trips.js
```diff
@@ -0,0 +1,29 @@
+import Http from '../http';
+import { TRIP_STATUS } from '../constants';
+
+class Trips {
+
+  static async updateTrip(tripOptions={}, status = TRIP_STATUS.UNKNOWN) {
```
kochis, line 6: Why have the default param in this case?

corypisano: ah yeah not needed

kochis:
You should also bump the version file (to match what's published) https://github.com/radarlabs/radar-sdk-js/blob/master/src/version.js

demo/test.html
```diff
@@ -137,6 +156,46 @@ <h3>Events</h3>
         });
       });
 
+      $("#trip-start-btn").click(function(e) {
```
jkao, line 159: hot take but i miss jquery

demo/test.html
```diff
@@ -137,6 +156,46 @@ <h3>Events</h3>
...
+        let destinationGeofenceExternalId = $('#trip-destination-externalId').val();
+
+        const tripOptions = {
+          externalId,
+          mode,
+          destinationGeofenceTag,
+          destinationGeofenceExternalId,
+        }
+
+        Radar.startTrip(tripOptions, function(err, responseObj, response) {
```
jkao, line 173: I know this is how nick designed it, but I'm not a fan of these global static functions that operate on an individual user level. something something concurrency

jkao: it feels like `user.startTrip` may have been a more appropriate API

jkao: anyways, nothing actionable here

src/api/track.js
```diff
@@ -36,9 +41,12 @@ class Track {
       sdkVersion: SDK_VERSION,
       stopped: true,
       userId,
+      tripOptions,
     };
 
-    const response = await Http.request('POST', `v1/track`, body);
+    const trackEndpoint = Cookie.getCookie(Cookie.TRACK_ENDPOINT) || 'v1/track';
```
jkao, line 47: do we need this to be customizable? I was under the impression it'd only be relevant for the host

corypisano: yeah we're going to expose a new track endpoint under /app that will use session auth for mock tracking in the dashboard (diff rate limits, etc)

jkao: ah, good point

demo/test.html
corypisano: sometimes.. for quick 1 pager stuff yeah

src/constants.js
```diff
@@ -0,0 +1,9 @@
+export const TRIP_STATUS = {
```
jsani-radar, line 1: can we name this file something a bit more descriptive like we do our other const files? https://github.com/radarlabs/radar-sdk-js/blob/master/src/error_codes.js

src/http.js
```diff
@@ -44,11 +44,21 @@ class Http {
...
       }
 
+      // set headers
       xhr.setRequestHeader('Authorization', publishableKey);
       xhr.setRequestHeader('Content-Type', 'application/json');
       xhr.setRequestHeader('X-Radar-Device-Type', 'Web');
       xhr.setRequestHeader('X-Radar-SDK-Version', SDK_VERSION);
 
+      // set custom headers if present
+      const customHeaders = Cookie.getCookie(Cookie.CUSTOM_HEADERS);
```
jsani-radar, line 54: does this need to be stored in cookies? or could it be passed through?

jkao (dismissed):

src/http.js
corypisano: hm passed through how? with each API request?

jsani-radar: yeah guess so, it's prob fine since the custom headers will be the same across requests

src/index.js
```diff
@@ -62,6 +65,15 @@ class Radar {
     Cookie.setCookie(Cookie.USER_ID, String(userId).trim());
   }
 
+  static setDeviceId(deviceId) {
```
jkao, line 68: what's a little funky is we also have a `Device` class that kind of does these things. I think we don't need to worry about it for now since this is the pattern we have for most of the SDKs, but we should prob take a pass on where similar domain things should be colocated.

jkao (approved):

src/index.js
corypisano: agreed

PR #46, opened 2021-07-01
v3.1.0 Matrix API

corypisano:
Add Matrix API support
Add `events` to update trip results
Add config for base api path


corypisano:
https://app.clubhouse.io/radarlabs/story/6528/return-events-in-js-sdk-patch-trips

src/api/matrix.js
```diff
@@ -0,0 +1,37 @@
+import Http from '../http';
+import Navigator from '../navigator';
+
+class Matrix {
+  static async getMatrixDistances(routingOptions={}) {
```
kochis, line 5: Should this just be in `src/api/routing.js`? Not sure we need a separate file for matrix

corypisano: ah yea i'll move it

andlum (approved):
The changes look good to me; I didn't know how to QA it but I mostly just cross-referenced the [iOS docs](https://radar.com/documentation/sdk/ios#matrix) for the Matrix API support

PR #48, opened 2022-01-05
Update package.json

nickpatrick:
- .io to .com
- consistency with other platforms


kochis:
I think we need to update node in CircleCI

PR #52, opened 2022-03-01
Add Matrix API

samfader:
This PR adds the Matrix API to the SDK. It also adds some notes to the README on how to spin up the SDK and run/test it locally. Finally, it adds a bit of UI to the index.html page to test out the Matrix capabilities.

cc @brettguenther

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,12 @@ See the Radar overview documentation [here](https://radar.io/documentation).
 
 Then, see the Radar web SDK documentation [here](https://radar.io/documentation/sdk/web).
 
+## Running locally
+
+1. Clone the repo
+2. `npm install`
+3. To run the SDK in a browser, run `npm run build && node demo/server.js`, which should open a tab in your browser of choice. Modify the `/src/index.html` file as needed
+
 ## Support
 
 Have questions? We're here to help! Email us at [support@radar.io](mailto:support@radar.io).
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Then, see the Radar web SDK documentation [here](https://radar.io/documentation/
 1. Clone the repo
 2. `npm install`
 3. To run the SDK in a browser, run `npm run build && node demo/server.js`, which should open a tab in your browser of choice. Modify the `/src/index.html` file as needed
+4. To run tests, run `npm run test`
 
 ## Support
 
--- a/demo/test.html
+++ b/demo/test.html
@@ -61,6 +61,30 @@
       <!-- By Sam Herbert (@sherb), for everyone. More @ http://goo.gl/7AJzbL -->
       <svg width="44" height="44" viewBox="0 0 44 44" xmlns="http://www.w3.org/2000/svg" stroke="#007bff">
         <g fill="none" fill-rule="evenodd" stroke-width="2">
+  <div class="pl-3 pt-3">
+    <a href="javascript:void(0)" id="show-matrix-options">Show Matrix API options</a>
+    <form id="matrix-form" class="form mb-0 d-none">
+      <input id="matrix-origins" class="form-control mr-sm-2 mb-1" type="text"
+        placeholder="Enter pipe-delimited origin coordinates (latitude0,longitude0|latitude1,longitude1|....)"
+        style="width: 400px;" />
+      <input id="matrix-destinations" class="form-control mr-sm-2 mb-1" type="text"
+        placeholder="Enter pipe-delimited destination coordinates (latitude0,longitude0|latitude1,longitude1|....)"
+        style="width: 400px;" />
+      <select id="matrix-mode" class="form-control mr-sm-2 mb-1" style="width: 400px;">
+        <option key="" disabled>Select Mode</option>
+        <option key="car">car</option>
+        <option key="foot">foot</option>
+        <option key="bike">bike</option>
+      </select>
+      <select id="matrix-units" class="form-control mr-sm-2 mb-1" style="width: 400px;">
+        <option key="" disabled>Select Units</option>
+        <option key="metric">metric</option>
+        <option key="imperial">imperial</option>
+      </select> 
+      <button id="matrix-run-btn" class="btn btn-primary">Perform Request</button>
+    </form>
+  </div>
+
         <circle cx="22" cy="22" r="1">
           <animate attributeName="r" begin="0s" dur="1.8s" values="1; 20" calcMode="spline" keyTimes="0; 1" keySplines="0.165, 0.84, 0.44, 1" repeatCount="indefinite" />
           <animate attributeName="stroke-opacity" begin="0s" dur="1.8s" values="1; 0" calcMode="spline" keyTimes="0; 1" keySplines="0.3, 0.61, 0.355, 1" repeatCount="indefinite" />
@@ -75,8 +99,15 @@
 
     <div class="error p-3 d-none">
       <div id="error-message" class="alert alert-danger" role="alert">
+  <div class="matrix-responses p-3 d-none">
+    <div>
+      <h3>Matrix Response</h3>
+      <div class="matrix-response">
+        <pre><code id="matrix"></code></pre>
       </div>
     </div>
+  </div>
+
 
     <div class="results p-3 d-none">
       <div>
@@ -194,10 +225,40 @@
           $('#show-trip-options').text('Show trip options');
           $('#trip-form').addClass('d-none');
         }
+    $("#matrix-run-btn").click(function (e) {
+      $('.matrix-responses').addClass('d-none');
+      e.preventDefault();
+      let origins = $('#matrix-origins').val();
+      let destinations = $('#matrix-destinations').val();
+      let mode = $('#matrix-mode').val();
+      let units = $('#matrix-units').val();
+
+      const matrixOptions = {
+        origins,
+        destinations,
+        mode,
+        units,
+      }
+
+      Radar.getMatrix(matrixOptions, function (err, responseObj, response) {
+        const trip = responseObj.trip;
+        $('#matrix').html(formatCode(response));
+
+        $('.matrix-responses').removeClass('d-none');
+        console.log(err, trip, response);
       });
+    });
 
       function formatCode(object) {
         return JSON.stringify(object, null, '  ');
+    $("#show-matrix-options").click(function (e) {
+      e.preventDefault();
+      if ($("#matrix-form").hasClass("d-none")) {
+        $('#show-matrix-options').text('Hide Matrix API options');
+        $('#matrix-form').removeClass('d-none');
+      } else {
+        $('#show-matrix-options').text('Show Matrix API options');
+        $('#matrix-form').addClass('d-none');
       }
     </script>
   </body>
--- a/demo/test.html
+++ b/demo/test.html
@@ -1,66 +1,70 @@
 <html>
-  <head>
-    <title>Radar | Test SDK Script</title>
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLPMO" crossorigin="anonymous">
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -12,6 +12,13 @@ See the Radar overview documentation [here](https://radar.io/documentation).
 
 Then, see the Radar web SDK documentation [here](https://radar.io/documentation/sdk/web).
 
+## Running locally
```
jsani-radar, line 15: incredible, thank you 🙇🏽‍♂️

demo/test.html
```diff
@@ -1,204 +1,287 @@
 <html>
-  <head>
-    <title>Radar | Test SDK Script</title>
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.1.3/css/bootstrap.min.css" integrity="sha384-MCw98/SFnGE8fJT3GXwEOngsV7Zt27NXFoaoApmYm81iuXoPkFOJwJ8ERdknLPMO" crossorigin="anonymous">
-    <script src="https://code.jquery.com/jquery-3.3.1.slim.min.js" integrity="sha256-3edrmyuQ0w65f8gfBsqowzjJe2iM6n0nKciPUp8y+7E=" crossorigin="anonymous"></script>
 
-    <script src="../dist/radar.js"></script>
+<head>
```
jsani-radar, line 3: I don't think this is a crazy formatting change - basically, it looks like the `<head>` tag is on the same level as the parent `<html>` tag, when it should be 2 spaces over. Should be fine after that as this mostly looks like whitespace changes

demo/test.html
```diff
@@ -1,204 +1,287 @@
...
+  </div>
 
-    <div class="loader pl-3 d-none">
-      <!-- By Sam Herbert (@sherb), for everyone. More @ http://goo.gl/7AJzbL -->
-      <svg width="44" height="44" viewBox="0 0 44 44" xmlns="http://www.w3.org/2000/svg" stroke="#007bff">
-        <g fill="none" fill-rule="evenodd" stroke-width="2">
+  <div class="pl-3 pt-3">
+    <a href="javascript:void(0)" id="show-matrix-options">Show Matrix API options</a>
+    <form id="matrix-form" class="form mb-0 d-none">
+      <input id="matrix-origins" class="form-control mr-sm-2 mb-1" type="text"
```
jsani-radar, line 71: @corypisano do you think this may be suited better on our API demo page? that's powered by our Web SDK today, so once we have this method we'd be able to layer a UI on top

demo/test.html
```diff
@@ -1,204 +1,287 @@
...
-    </script>
-  </body>
-</html>
+    });
+
+    function formatCode(object) {
+      return JSON.stringify(object, null, '  ');
+    }
+  </script>
+</body>
```
jsani-radar, line 285: ditto here, the body closing tag is at the same level as the html one

src/api/matrix.js
```diff
@@ -0,0 +1,36 @@
+import Http from '../http';
+import Navigator from '../navigator';
+
+class Matrix {
+  static async getMatrixDistances(routingOptions={}) {
+    if (!routingOptions.origins) {
+      const { latitude, longitude } = await Navigator.getCurrentPosition();
+      routingOptions.origins = { latitude, longitude };
```
jsani-radar, line 8: I think maybe this should be an array? since the matrix API takes an array of coordinates

src/api/matrix.js
```diff
@@ -0,0 +1,36 @@
...
+
+    let {
+      origins,
+      destinations,
+      mode,
+      units,
+    } = routingOptions;
+
+    origins = origins;
+    mode = mode;
```
jsani-radar, line 19: don't think these two lines are necessary, you're already destructuring these vars from `routingOptions`

src/api/matrix.js
```diff
@@ -0,0 +1,36 @@
...
+
+    if (destinations) {
+      destinations = destinations;
+    }
```
jsani-radar, line 23: ditto here, the matrix API requires destinations

test/api/matrix.test.js
```diff
@@ -0,0 +1,78 @@
+const chai = require('chai');
```
jsani-radar, line 1: think we can slot this test in `routing.test.js` as we have it nested as a Routing API in our docs

jsani-radar (commented):
awesome first contribution! looking really solid

few small suggested fixes and questions for my knowledge

src/api/matrix.js
samfader: @jsani-radar I think the Matrix API just takes a pipe-delimited string: https://radar.com/documentation/api#matrix

> origins (string, required): A list of origins. A pipe-delimited string in the format latitude0,longitude0|latitude1,longitude1|....

test/api/matrix.test.js
samfader: @jsani-radar would the best way to do this be a separate `describe('Matrix', ())` block in the routing.test.js test suite? or should it be part of the Routing block? the origin and destinations in the Matrix API are formatted a bit differently so I'd have to create separate variables if the latter

src/api/matrix.js
samfader: oh, wait, I see what you're saying...let me look into that

test/api/matrix.test.js
samfader: @jsani-radar went ahead and moved it into routing.test.js as described above, and all tests are passing

src/api/matrix.js
jsani-radar: yup makes sense - though I'm not seeing where this is transformed to a pipe-delimited string, as it seems like we're passing the object arg through to the API

demo/test.html
jsani-radar: bumping this - did you have a chance to try tabbing over everything within `<head>` and `<body>` 2 spaces?

test/api/routing.test.js
```diff
@@ -82,3 +83,67 @@ describe('Routing', () => {
     });
   });
 });
+
+describe('Matrix', () => {
```
jsani-radar, line 87: looking at our other test files with multiple endpoints (geocoding, search), it seems like there's one top-level `describe` block, with each API nested as it's own `context`, see the following examples from search:
1. [searchPlaces](https://github.com/radarlabs/radar-sdk-js/blob/master/test/api/search.test.js#L44)
2. [searchGeofences](https://github.com/radarlabs/radar-sdk-js/blob/master/test/api/search.test.js#L83)
3. [autocomplete](https://github.com/radarlabs/radar-sdk-js/blob/master/test/api/search.test.js#L122)

could we follow the same pattern here and nest a `getMatrixDistances` context within the top-level `Routing` block?

apologies about being super nitpicky here, it's just easier for a new dev to reason about the tests if they follow the same structure

test/api/routing.test.js
```diff
@@ -82,3 +83,67 @@ describe('Routing', () => {
...
+
+          return Matrix.getMatrixDistances({ destination, mode, units })
+            .then((response) => {
+              expect(response).to.equal(routingResponse);
+            });
+        });
+      });
+    });
+  });
+});
```
jsani-radar, line 149: nit: `\n`

jsani-radar:
as I mentioned in Slack, let's try bumping the circleci image to `12.22.7`

test/api/routing.test.js
samfader: nope, this makes sense! thanks @jsani-radar - updated

samfader:
Thanks @jsani-radar, Circle tests passing now. Made some updates, and have one more question for you inline.

src/api/matrix.js
jsani-radar: what I mean here is that we need to transform the `origins` var to the format that the API expects, see this example for what we do in our `searchPlaces` call with chains / categories / groups - https://github.com/radarlabs/radar-sdk-js/blob/master/src/api/search.js#L22-L30

they're [passed in as arrays](https://github.com/radarlabs/radar-sdk-js/blob/master/test/api/search.test.js#L21-L23), but then formatted correctly

demo/test.html
corypisano: im good having it here as well, for internal use vs developer example

src/api/matrix.js
corypisano: yeah +1, in this spot we can set origins to be an array like:

```
const { latitude, longitude } = await Navigator.getCurrentPosition();
routingOptions.origins = [{ latitude, longitude }];
```

test/api/routing.test.js
```diff
@@ -81,4 +87,44 @@ describe('Routing', () => {
...
+            });
+        });
+      });
+
+      describe('all args given', () => {
+        it('should return a routing response', () => {
+          navigatorStub.resolves(matrixOrigin);
+          httpStub.resolves(matrixResponse);
+
+          return Matrix.getMatrixDistances({ matrixDestination, matrixMode, units })
```
corypisano, line 122: ah these don't look like the right input names

`{ matrixDestination, matrixMode, units }`

the actual names are `{ origins,  destinations, mode, units }`
can do `getMatrixDistances({ destinations: MatrixDestination, mode: matrixMode })` if the variable name doesn't align with the input

src/api/matrix.js
```diff
@@ -5,7 +5,7 @@ class Matrix {
   static async getMatrixDistances(routingOptions={}) {
     if (!routingOptions.origins) {
       const { latitude, longitude } = await Navigator.getCurrentPosition();
-      routingOptions.origins = { latitude, longitude };
+      routingOptions.origins = [{ latitude, longitude }, { latitude, longitude }];
```
jsani-radar, line 8: looks like this is duplicating the origin twice?

samfader:
@corypisano @jsani-radar I believe this is ready for one more round of reviews. Thanks for all the assistance!

src/api/matrix.js
```diff
@@ -0,0 +1,33 @@
...
+    let {
+      origins,
+      destinations,
+      mode,
+      units,
+    } = routingOptions;
+
+    origins = origins.map(origin => { return `${origin.latitude}, ${origin.longitude}` }).join('|')
+    destinations = destinations || []
+    destinations.map(destination => { return `${destination.latitude}, ${destination.longitude}` }).join('|')
```
jsani-radar, line 20: not sure why our linter didn't catch this but we should have `;` on the end of these statements

src/api/matrix.js
```diff
@@ -0,0 +1,33 @@
...
+
+    let {
+      origins,
+      destinations,
+      mode,
+      units,
+    } = routingOptions;
+
+    origins = origins.map(origin => { return `${origin.latitude}, ${origin.longitude}` }).join('|')
+    destinations = destinations || []
```
jsani-radar, line 19: should prob do this for `origins` too just in case

test/api/routing.test.js
```diff
@@ -32,6 +33,18 @@ describe('Routing', () => {
 
   const routingResponse = { meta: {}, routes: {} };
 
+  const matrixMode = 'car';
+  // const matrixOrigin = [{
+  //   latitude: 40.70390, longitude: }];
```
jsani-radar, line 38: cleanup?

test/api/routing.test.js
```diff
@@ -32,6 +33,18 @@ describe('Routing', () => {
...
+  const matrixOrigin = [{ 
+    latitude: 40.70390, 
+    longitude: -73.98690}];
```
jsani-radar, line 41: yeah something seems off with the formatting here /cc @corypisano

test/api/routing.test.js
jsani-radar: bump - let's add a new line here for consistency

test/api/routing.test.js
corypisano: looks like we need to configure lint and run in CI. (get on par with frontend setup). will add a shortcut ticket

corypisano: for now, can format like below so it matches `matrixDestination`
```
const matrixOrigin = [
  {latitude: 40.70390,  longitude: -73.98690}
];
```

corypisano (approved):
Looks good to me 👍  just the nit on formatting

PR #56, opened 2022-03-11
Matrix hotfix

samfader:
Fix to the getMatrix method to ensure the correct top level response objects are returned. 

cc @kochis @corypisano @jsani-radar


kochis:
Added `matrix` api call to demo page

image

PR #63, opened 2022-06-03
Es and ts

moczix:
i created es build output + added some typings


kochis:
ES build & typings are now included as of [v4](https://github.com/radarlabs/radar-sdk-js/releases/tag/v4.0.0)

PR #69, opened 2022-07-14
Add a Radar.clearTripOptions interface

jkao:


corypisano (approved):

src/index.js
```diff
@@ -197,20 +195,24 @@ class Radar {
...
       })
       .catch(handleError(callback));
   }
 
   static setTripOptions(tripOptions) {
     if (!tripOptions) {
+      Radar.clearTripOptions();
       return;
     }
     Storage.setItem(Storage.TRIP_OPTIONS, JSON.stringify(tripOptions));
```
jrtibbetts, line 209: Would 

```
if (!tripOptions) {
    Storage.setItem(Storage.TRIP_OPTIONS, JSON.stringify(tripOptions));
}
```

have the same effect, or does explicitly calling `Storage.removeItem()` have some side effects that need to happen?

PR #86, opened 2023-01-26
RAD-5740: cache-recent-locations-in-websdk-with-variable-ttl

boonboonsiri:
### Requirements
- [x] Linear ticked is associated with PR
- [x] Tests have been added (if not, explain why)
- [ ] Feature flags have been added and default to off in production
- [x] [Docs](https://github.com/radarlabs/documentation/tree/main/docs) have been updated (if applicable)
- [ ] Internal docs have been updated
- [x] QA is done and documented below

Difficult to write test for. Tested manually:

### QA details
*Include screenshots, responses, or steps to verify the code works as expected*

Can verify locally by setting up local sdk, or verify that it works using vercel:

https://frontend-git-boon-sdk-demo-branch-radarlabs.vercel.app/demo/js

Currently cached location time is 30 seconds. So go to link in safari, enter prompt for location.

After you can continuously reload for up to 30 seconds without being re-prompted for your location, but then after 30 seconds you'll get prompted again

I probably should have just installed loom on safari but here we are:

https://youtu.be/VU5nAQ74pyo

docs:
https://github.com/radarlabs/documentation/pull/232

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -19,7 +19,7 @@ const defaultCallback = () => {};
 
 const handleError = (callback) => {
   return (err) => {
-
+    
     // Radar Error
     if (typeof err === 'string') {
       callback(err, {});
@@ -46,11 +46,19 @@ class Radar {
     return STATUS;
   }
 
-  static initialize(publishableKey) {
+  static initialize(publishableKey, options={}) {
     if (!publishableKey) {
       console.error('Radar "initialize" was called without a publishable key');
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
+
+    if(options){
+      let {timeToLive} = options
+
+      if(timeToLive && typeof(timeToLive) === 'number'){
+        Storage.setItem(Storage.LOCATION_TIME_TO_LIVE, timeToLive)
+      }
+    }
   }
 
   static setHost(host, baseApiPath = API_VERSION) {
--- a/src/index.js
+++ b/src/index.js
@@ -52,16 +52,15 @@ class Radar {
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
 
-    if(Object.keys(options).length === 0){
-      const {locationTimeToLive} = options
 
-      if (locationTimeToLive) {
-        const number = Number(locationTimeToLive);
-        if (Number.isNaN(number)) {
-          console.warn('Radar SDK: invalid number for option "locationTimeToLive"');
-        } else {
-          Storage.setItem(Storage.LOCATION_TIME_TO_LIVE, locationTimeToLive);
-        }
+    const {cacheLocationMinutes} = options
+
+    if (cacheLocationMinutes) {
+      const number = Number(cacheLocationMinutes);
+      if (Number.isNaN(number)) {
+        console.warn('Radar SDK: invalid number for option "cacheLocationMinutes"');
+      } else {
+        Storage.setItem(Storage.CACHE_LOCATION_MINUTES, cacheLocationMinutes);
       }
     }
   }
--- a/src/index.js
+++ b/src/index.js
@@ -51,7 +51,6 @@ class Radar {
       console.error('Radar "initialize" was called without a publishable key');
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
-
     if(options){
       let {timeToLive} = options
 
--- a/src/index.js
+++ b/src/index.js
@@ -52,8 +52,7 @@ class Radar {
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
 
-
-    const {cacheLocationMinutes} = options
+    const { cacheLocationMinutes } = options;
 
     if (cacheLocationMinutes) {
       const number = Number(cacheLocationMinutes);
@@ -62,6 +61,8 @@ class Radar {
       } else {
         Storage.setItem(Storage.CACHE_LOCATION_MINUTES, cacheLocationMinutes);
       }
+    }else{
+      Storage.removeItem(Storage.CACHE_LOCATION_MINUTES);
     }
   }
 
--- a/src/index.js
+++ b/src/index.js
@@ -61,12 +61,12 @@ class Radar {
       } else {
         Storage.setItem(Storage.CACHE_LOCATION_MINUTES, cacheLocationMinutes);
       }
-    }else{
+    } else {
       Storage.removeItem(Storage.CACHE_LOCATION_MINUTES);
     }
   }
 
-  static setHost(host, baseApiPath = API_VERSION) {
+  static setHost(host, baseApiPath=API_VERSION) {
     Storage.setItem(Storage.HOST, host);
     Storage.setItem(Storage.BASE_API_PATH, baseApiPath);
   }
@@ -163,7 +163,6 @@ class Radar {
 
     if (typeof arg0 === 'function') {
       callback = arg0;
-
     } else {
       location = arg0;
       callback = arg1;
@@ -293,7 +292,6 @@ class Radar {
 
     if (typeof arg0 === 'function') {
       callback = arg0;
-
     } else if (typeof arg0 === 'object') {
       geocodeOptions = arg0;
       callback = arg1;
--- a/src/index.js
+++ b/src/index.js
@@ -19,7 +19,7 @@ const defaultCallback = () => {};
 
 const handleError = (callback) => {
   return (err) => {
-    
+
     // Radar Error
     if (typeof err === 'string') {
       callback(err, {});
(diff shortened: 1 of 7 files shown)


boonboonsiri:
Ask about which branch to merge this into

boonboonsiri:
Note time to live specifically exists for `Track.trackOnce` and not for all location request. Not that difficult to set this up for all location request (just need to move code)

src/index.js
```diff
@@ -46,11 +46,18 @@ class Radar {
...
   }
 
-  static initialize(publishableKey) {
+  static initialize(publishableKey, options={}) {
     if (!publishableKey) {
       console.error('Radar "initialize" was called without a publishable key');
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
+    if(options){
+      let {locationTimeToLive} = options
```
kochis, line 55: use `const` instead of `let` if it's not being reassigned.

src/index.js
```diff
@@ -46,11 +46,18 @@ class Radar {
...
+
+      if(locationTimeToLive && typeof(locationTimeToLive) === 'number'){
```
kochis, line 57: Maybe we should throw a warning if it's set but not a number, but also be more lenient and accept number strings.
```js
if (locationTimeToLive) {
  const number = Number(locationTimeToLive);
  if (Number.isNaN(number)) {
    console.warn('Radar SDK: invalid number for option "locationTimeToLive"'));
  } else {
    Storage.setItem(Storage.LOCATION_TIME_TO_LIVE, locationTimeToLive)
  }
}
```

src/navigator.js
```diff
@@ -17,6 +32,11 @@ class Navigator {
 
           const { latitude, longitude, accuracy } = position.coords;
 
+          Storage.setItem(Storage.LAST_LOCATION_TIME, Date.now())
+          Storage.setItem(Storage.LATITUDE, latitude)
+          Storage.setItem(Storage.LONGITUDE, longitude)
+          Storage.setItem(Storage.ACCURACY, accuracy)
```
kochis, line 38: I think we can just roll these all up into JSON and use one key instead of having the separate fields. Also maybe we should only cache it if the setting is enabled.
```js
if (locationTimeToLive) {
  const lastLocation = { latitude, longitude, accuracy, updatedAt: Date.now() };
  Storage.setItem(Storage.LAST_LOCATION, JSON.stringify(lastLocation));
}
```
And then when we read it out, we can do `JSON.parse`
```js
if (locationTimeToLive) {
  try {
    const lastLocation = JSON.parse(Storage.getItem(Storage.LAST_LOCATION));
    const { latitude, longitude, accuracy, updatedAt } = lastLocation;
    // check expiration stuff goes here

  } catch (e) {
    console.warn('Radar SDK: could not load cached location.');
  }
  const 
}

File: src/index.js (hunk 1 above)
Comment (boonboonsiri) [reply src/index.js:55]: 👍

File: src/index.js [hunk 4]
```diff
@@ -46,11 +46,18 @@ class Radar {
     return STATUS;
   }
 
-  static initialize(publishableKey) {
+  static initialize(publishableKey, options={}) {
     if (!publishableKey) {
       console.error('Radar "initialize" was called without a publishable key');
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
+    if(options){
```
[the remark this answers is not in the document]
boonboonsiri: 👍

src/index.js
boonboonsiri: 👍

boonboonsiri:
Okay changes made. Back at you @kochis

src/navigator.js
```diff
@@ -17,6 +36,11 @@ class Navigator {
 
           const { latitude, longitude, accuracy } = position.coords;
 
+          if (locationTimeToLive) {
+            const lastLocation = { latitude, longitude, accuracy, updatedAt: Date.now() };
+            Storage.setItem(Storage.LAST_LOCATION, JSON.stringify(lastLocation));
+          }
```
kochis, line 42: It probably makes sense to bake in the expiration here too.
```js
          if (locationTimeToLive) {
            const updatedAt = Date.now();
            const expiresAt = updatedAt + (locationTimeToLive * 60 * 1000); // convert to ms
            const lastLocation = { latitude, longitude, accuracy, updatedAt, expiresAt };
            Storage.setItem(Storage.LAST_LOCATION, JSON.stringify(lastLocation));
          }
```
Then when we read it out, we just need to check
```js
if (Date.now() < parseInt(lastLocation.expiresAt())) {
```

src/index.js
```diff
@@ -46,11 +46,24 @@ class Radar {
...
 
-  static initialize(publishableKey) {
+  static initialize(publishableKey, options={}) {
     if (!publishableKey) {
       console.error('Radar "initialize" was called without a publishable key');
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
+
+    if(Object.keys(options).length === 0){
+      const {locationTimeToLive} = options
```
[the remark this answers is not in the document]
boonboonsiri: 👍

src/index.js
```diff
@@ -46,11 +46,24 @@ class Radar {
...
   }
 
-  static initialize(publishableKey) {
+  static initialize(publishableKey, options={}) {
     if (!publishableKey) {
       console.error('Radar "initialize" was called without a publishable key');
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
+
+    if(Object.keys(options).length === 0){
```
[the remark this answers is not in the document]
boonboonsiri: image
https://stackoverflow.com/questions/6072590/how-to-match-an-empty-dictionary-in-javascript

That's apparently one of the ways to handle empty dictionary's in JS

src/navigator.js
```diff
@@ -8,6 +9,24 @@ class Navigator {
         return reject(STATUS.ERROR_LOCATION);
       }
 
+      let locationTimeToLive = parseFloat(Storage.getItem(Storage.LOCATION_TIME_TO_LIVE));
```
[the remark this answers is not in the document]
boonboonsiri: 👍

src/index.js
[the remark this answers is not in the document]
boonboonsiri: Wait I see what you mean about it being 0. Which is really weird cause when I was testing it yesterday the code worked, but maybe it was just using old cache values and I didn't realize. Also yeah that's a good point that this is unneeded

src/index.js
```diff
@@ -46,11 +46,23 @@ class Radar {
...
 
-  static initialize(publishableKey) {
+  static initialize(publishableKey, options={}) {
     if (!publishableKey) {
       console.error('Radar "initialize" was called without a publishable key');
     }
     Storage.setItem(Storage.PUBLISHABLE_KEY, publishableKey);
+
+
+    const {cacheLocationMinutes} = options
```
corypisano, line 56: nit: space around the brackets `{ cacheLocationMinutes }` for consistency, and semicolon at end

src/index.js
```diff
@@ -46,11 +46,23 @@ class Radar {
...
+
+    if (cacheLocationMinutes) {
```
corypisano, line 58: hm we might want an `else` that clears it if `cachedLocationMinutes` is not passed in. in the past, I've run into issues from local storage items getting set and sticking around unintended

for example if you call `initialize` with cachedLocationMinutes=5, then you want to turn it off or on a separate initialize it will stick around even if you don't pass in the field.

src/index.js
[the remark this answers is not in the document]
boonboonsiri: 👍

src/navigator.js
```diff
@@ -8,6 +9,21 @@ class Navigator {
         return reject(STATUS.ERROR_LOCATION);
       }
 
+      const cacheLocationMinutes = parseFloat(Storage.getItem(Storage.CACHE_LOCATION_MINUTES));
+      if (cacheLocationMinutes) {
+        try {
+          const lastLocation = JSON.parse(Storage.getItem(Storage.LAST_LOCATION));
```
corypisano, line 15: we should probably skip the rest if LAST_LOCATION is null, think it will error for the first call and go to the warn since there will be undefined fields

src/index.js
boonboonsiri: 👍

src/index.js
boonboonsiri: 👍

src/navigator.js
boonboonsiri: 👍

boonboonsiri:
Back at you @corypisano

corypisano:
cool i'll bump the sdk version, and just gonna fix some formatting (we should bring lint setup here but separate issue)

corypisano (approved):

PR #88, opened 2023-02-17
remove geocoding param

tjulien:


src/index.js
```diff
@@ -286,18 +286,14 @@ class Radar {
       .catch(handleError(callback));
   }
 
-  static ipGeocode(arg0, arg1=defaultCallback) {
```
kochis, line 289: I think this is technically a breaking change. 

It's probably best to leave the initial arg in place, but do a `console.warn` in the case where `arg0` as an object is provided.
```js
console.warn('Radar SDK: ipGeocode parameters have been deprecated.');
```

tjulien: updated, can you take a look to see if it's what you had in mind?

kochis: Yep, looks good.

tjulien: Cool. Good to merge?

PR #89, opened 2023-03-07
Fix for Http.request() method required data param

JoRoRadar:


kochis:
In addition to this fix, lets also set a default argument for `data` in the Http class: https://github.com/radarlabs/radar-sdk-js/blob/030d05ef6eb0fcc9109ad397dc6a32addacfe409/src/http.js#L9

```js
static request(method, path, data={}) {
```

kochis (approved):

PR #90, opened 2023-03-09
Cory/inf 627 address validation web sdk

corypisano:
- Add `radar.ValidateAddress()` method 
- Add `expandUnits` as a param to `radar.autocomplete()`

QA: 

validateAddress() response:
Screen Shot 2023-03-09 at 5 46 42 PM

### DIFF
--- a/src/api/addresses.js
+++ b/src/api/addresses.js
@@ -9,7 +9,8 @@ class Addresses {
       number,
       postalCode,
       street,
-      unit // optional
+      unit,
+      addressLabel,
     } = addressOptions;
 
     const params = {
@@ -19,7 +20,8 @@ class Addresses {
       number,
       postalCode,
       street,
-      unit
+      unit,
+      addressLabel,
     };
 
     return Http.request('GET', 'addresses/validate', params);
--- a/src/api/addresses.js
+++ b/src/api/addresses.js
@@ -1,0 +1,29 @@
+import Http from '../http';
+
+class Addresses {
+  static async validateAddress(addressOptions={}) {
+    let {
+      country,
+      state,
+      city,
+      number,
+      postalCode,
+      street,
+      unit // optional
+    } = addressOptions;
+
+    const params = {
+      country,
+      state,
+      city,
+      number,
+      postalCode,
+      street,
+      unit
+    };
+
+    return Http.request('GET', 'addresses/validate', params);
+  }
+}
+
+export default Addresses;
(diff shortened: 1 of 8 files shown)


test/api/addresses.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const country = 'US';
+  const state = 'NY';
+  const city = 'New York';
+  const number = '841';
+  const postalCode = '10010';
+  const street = 'Broadway';
+  const unit = '7';
+
+  const autocompleteResponse = { meta: {}, address: {} };
```
bschoeneweis, line 23: nit: `validateResponse`

bschoeneweis (approved):
Looks good 👍

src/api/addresses.js
```diff
@@ -0,0 +1,29 @@
+import Http from '../http';
+
+class Addresses {
+  static async validateAddress(addressOptions={}) {
+    let {
```
kochis, line 5: should this be `const`

corypisano: ye

tjulien (dismissed):

src/api/addresses.js
```diff
@@ -0,0 +1,29 @@
+import Http from '../http';
+
+class Addresses {
+  static async validateAddress(addressOptions={}) {
+    const {
+      country,
+      state,
```
bschoeneweis, line 7: changed to `countryCode` and `stateCode`

bschoeneweis (changes requested):

src/api/search.js
```diff
@@ -89,6 +89,7 @@ class Search {
       limit,
       layers,
       country,
```
bschoeneweis, line 91: We can also update to `countryCode` here (will accept in the endpoint now)

corypisano: ok yea i guess we can support both so nothing breaks

bschoeneweis (dismissed):
Cool will leave to you whether or not to update `/search/autocomplete`, probably good to follow suite w/ iOS and android

corypisano:
image

src/api/addresses.js
```diff
@@ -0,0 +1,29 @@
...
+class Addresses {
+  static async validateAddress(addressOptions={}) {
+    const {
+      countryCode,
+      stateCode,
+      city,
+      number,
+      postalCode,
+      street,
+      unit // optional
```
bschoeneweis, line 12: also will want to add `addressLabel`.  not sure if we care to do any validation in the sdk for param combos or leave it to the api

lmeier (approved):
👍

PR #91, opened 2023-03-14
bump version 3.7.0

corypisano:


lmeier (approved):
Correct that we want 3.6.2 instead of 3.7.0? If so, lgtm

lmeier (approved):

tjulien (approved):

PR #92, opened 2023-03-27
CORE-295 conversions to web sdk

david-goodfellow:
There is no idea of a "session" for web sdk so we will not be doing `opened_app` here. Instead we are just making the header changes for `logConversion` with the additional capability of taking in a `revenue` param

Open Q: should we add a migration guide file?

QA:
* added tests that are passing in the CI
* added to the local browser test and it is working
image
image

### DIFF
--- a/demo/test.html
+++ b/demo/test.html
@@ -39,8 +39,14 @@
     </div>
 
     <div class="pl-3 pt-3">
-      <form id="custom-event" class="form-inline mb-0">
-        <button id="custom-event-btn" class="btn btn-primary">Send Custom Event</button>
+      <form id="conversion-event" class="form-inline mb-0">
+        <button id="conversion-event-btn" class="btn btn-primary">Send Conversion Event</button>
+      </form>
+    </div>
+
+    <div class="pl-3 pt-3">
+      <form id="conversion-event" class="form-inline mb-0">
+        <button id="conversion-revenue-event-btn" class="btn btn-primary">Send Conversion Revenue Event</button>
       </form>
     </div>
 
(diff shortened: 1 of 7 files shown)


demo/test.html
```diff
@@ -39,8 +39,14 @@
     </div>
 
     <div class="pl-3 pt-3">
-      <form id="custom-event" class="form-inline mb-0">
-        <button id="custom-event-btn" class="btn btn-primary">Send Custom Event</button>
+      <form id="conversion-event" class="form-inline mb-0">
+        <button id="conversion-event-btn" class="btn btn-primary">Send Conversion Event</button>
```
corypisano, line 43: nit: Log conversion event

package-lock.json
```diff
@@ -4021,98 +4021,128 @@
...
       "engines": {
         "node": ">=0.10.0"
       }
     },
     "node_modules/fsevents/node_modules/aproba": {
       "version": "1.2.0",
+      "dev": true,
       "inBundle": true,
-      "license": "ISC"
+      "license": "ISC",
```
corypisano, line 4043: hm looks like a lot of package-lock changes, not sure if desired or not.  we might just want to only touch `version` at the top

demo/test.html
```diff
@@ -215,13 +221,27 @@ <h3>Matrix</h3>
...
         e.preventDefault();
-        const customEventData = {
-          "type": "button_click",
+        const conversionEventData = {
+          "name": "purchase_click",
+          "revenue": 10,
           "metadata": {"page":"test"}
         }
-        Radar.sendEvent(customEventData, function(err, responseObj, response) {
+        console.log('here')
```
joeychik-radar, line 243: remove this?

joeychik-radar (approved):
one nit, looks good otherwise

corypisano (approved):

PR #94, opened 2023-03-28
Github Action for releases

tjulien:
Automate release process with Github Actions.

There are two workflows. Each one will publish the version to `npm`, and upload the artifacts to S3 for usage with script tags:
* `prereleased` - releases marked as pre-releases will be published with the `-beta` tag
* `released` - latest release

This also adds some checks to the release process:
* tags must use a `"v"` prefix (pre & production releases)
* pre-releases _must_ have a `-beta` tag associated with them
* production releases must not have any tag suffixes
* ensures the tag matches all version references in code

Also adds a utility script for updating version references in code `npm run bump-version -- <version>`
* If no version is provided, the rightmost value in incremented

### DIFF
--- a/.github/workflows/radar-release-actions.yml
+++ b/.github/workflows/radar-release-actions.yml
@@ -17,10 +17,10 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
       - uses: jakejarvis/s3-sync-action@master
-      env:
-        AWS_S3_BUCKET: js.radar.com
-        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-        AWS_REGION: 'us-east-1'
-        SOURCE_DIR: 'dist/'      # optional: defaults to entire repository
-        DEST_DIR: ${{ github.ref }}
+        env:
+          AWS_S3_BUCKET: js.radar.com
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: 'us-east-1'
+          SOURCE_DIR: 'dist/'      # optional: defaults to entire repository
+          DEST_DIR: ${{ github.ref }}
--- a/.github/workflows/radar-release-actions.yml
+++ b/.github/workflows/radar-release-actions.yml
@@ -26,4 +26,4 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_REGION: 'us-east-1'
           SOURCE_DIR: 'dist/'
-          DEST_DIR: ${{ github.ref }} # the release tag
+          DEST_DIR: ${{ github.ref_name }} # the release tag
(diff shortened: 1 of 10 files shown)


.github/workflows/radar-release-actions.yml
```diff
@@ -0,0 +1,26 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - uses: jakejarvis/s3-sync-action@master
+        env:
+          AWS_S3_BUCKET: js.radar.com
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: 'us-east-1'
+          SOURCE_DIR: 'dist/'      # optional: defaults to entire repository
+          DEST_DIR: ${{ github.ref }}
```
kochis, line 26: What's the value of this?

kochis:
The `dist/` files aren't checked in, so we probably need to include a build step in here (and probably tests too). 

We be good to also automate (or at least verify) that all of the versions match correctly:
* the release tag
* the [version file](https://github.com/radarlabs/radar-sdk-js/blob/master/src/version.js)
* the [package.json](https://github.com/radarlabs/radar-sdk-js/blob/master/package.json#L6)
* the [package-lock.json](https://github.com/radarlabs/radar-sdk-js/blob/master/package-lock.json#L9)

.github/workflows/radar-release-actions.yml
tjulien: It's the release tag.  I'll stick a `v` on the front

tjulien:
@kochis would be cool to cross check all of the versions in those files.  We don't really do that with the other SDKs at the moment, maybe can do that after we get this basic one working.  In the other notion pages we do document it, we can do that for the web sdk too (for example: https://www.notion.so/radarlabs/How-to-release-the-React-Native-SDK-21c70f16e0bd4fe28425923645cadda0)

tjulien:
@kochis I added steps for `npm build` and `npm test`

kochis:
Yeah, the steps to update all the files are also documented for JS, but if the point of this is to automate the build, it would be good to have this as well since it's part of the process.

We do have a test to ensure that they match: https://github.com/radarlabs/radar-sdk-js/blob/master/test/version.test.js

We don't need to automate the bumping of the versions, but would be good to have some additional step to ensure it matches the tag. I can take a stab at this (will just make a small npm script)

.github/workflows/radar-release-actions.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/checkout@v2
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm run build
+      - run: npm run test
+      - run: npm publish
```
corypisano, line 18: might need a way to support `npm publish --tag beta`

kochis: Yeah, good idea

tjulien: There are two different events we can use for release vs beta release:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

Note: The `prereleased` type will not trigger for pre-releases published from draft releases, but the `published` type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to `published` instead of `released` and `prereleased`.

(and then when creating the release, you check the box that says "this is a pre-release"

tjulien: Basically, `released` vs `prereleased`

kochis: Ah nice, will use that.

kochis:
Release uploaded to S3: https://js.radar.com/v3.7.1-beta.2/radar.js

Beta npm version published: https://www.npmjs.com/package/radar-sdk-js?activeTab=versions

PR #102, opened 2023-06-21
Add option to enable high accuracy

kochis:
Adds a new option `enableHighAccuracy` for [getCurrentPosition](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition).

## Usage
```js
Radar.initialize('prj_test_pk_...', { enableHighAccuracy: true });
```

## QA
**disabled (default)**
Screenshot 2023-06-21 at 2 36 54 PM

**enabled**
Screenshot 2023-06-21 at 2 36 03 PM

### DIFF
--- a/demo/test.html
+++ b/demo/test.html
@@ -138,7 +138,7 @@
 
         let publishableKey = $('#publishable-key').val();
 
-        Radar.initialize(publishableKey);
+        Radar.initialize(publishableKey, { enableHighAccuracy });
 
         Radar.trackOnce(function(err, responseObj, response) {
           const location = responseObj.location;
--- a/src/navigator.js
+++ b/src/navigator.js
@@ -26,6 +26,14 @@ class Navigator {
         }
       }
 
+      const enableHighAccuracy = Storage.getItem(Storage.ENABLE_HIGH_ACCURACY);
+
+      const positionOptions = {
+        maximumAge: 0,
+        timeout: 1000 * 30, // 30 seconds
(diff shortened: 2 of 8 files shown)


src/navigator.js
```diff
@@ -26,6 +26,14 @@ class Navigator {
         }
       }
 
+      const enableHighAccuracy = Storage.getItem(Storage.ENABLE_HIGH_ACCURACY);
```
jsani-radar, line 29: should this default to `true`? looks like it might be `false` if not specified

kochis: Probably makes sense to preserve the existing behavior (default to false).

demo/test.html
```diff
@@ -138,7 +138,7 @@ <h3>Matrix</h3>
 
         let publishableKey = $('#publishable-key').val();
 
-        Radar.initialize(publishableKey);
+        Radar.initialize(publishableKey, { enableHighAccuracy });
```
tjulien, line 141: The other SDKs pass in `desiredAccuracy = high, medium, or low` to `trackOnce()` or set it on `trackingOptions` for `starTracking()`.   So for consistency we probably want to do the same thing for `trackOnce()` in the js sdk.  Since it's a boolean at the browser level, maybe we map `high` to `true` and `medium` and `low` to `false`.

thx @jsani-radar for flagging in slack

tjulien: Since 4.x is going to be a major version bump, that should give us cover to introduce `trackOnce(options: {}, (err, responseObj, response))

PR #104, opened 2023-07-14
Radar SDK V4

kochis:
🚀 **New Features**
* SDK fully re-written in TypeScript with exposed type definitions
* Async functions now return a `Promise`
* Support for [Radar Maps](https://radar.com/documentation/maps/maps) powered by [MapLibre](https://maplibre.org/)
* Radar Autocomplete UI Kit 🆕

⚙️ **Breaking Changes**
* `radar-sdk-js` now has a peerDependency on the `maplibre-gl` package.
* Functions are no longer callback based (now return `Promise`)

See more in [MIGRATION.md](https://github.com/radarlabs/radar-sdk-js/blob/v4-beta/MIGRATION.md)


src/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      options,
+    );
+    Config.setup(radarOptions);
+
+    Logger.info(`initialized with ${live ? 'live' : 'test'} publishableKey.`);
+    if (options.debug) {
+      Logger.info(`using options: ${JSON.stringify(options)}`);
+    }
+
+    ConfigAPI.getConfig();
```
jaspk06, line 95: since `ConfigAPI.getConfig()` is a `Promise`, should we make `Radar.initialize` a `Promise` as well so that we can give the user the option to `await` for the initialization to complete? (noticed this when writing the tests)

something like 
```suggestion
  public static initialize(publishableKey: string, options: RadarOptions = {}): Promise<void> {
    if (!publishableKey) {
      throw new RadarPublishableKeyError('Publishable key required in initialization.');
    }

    if (isSecretKey(publishableKey)) {
      throw new RadarPublishableKeyError('Secret keys are not allowed. Please use your Radar publishable key.');
    }

    // store settings in global config
    const live = isLiveKey(publishableKey);
    const logLevel = live ? 'error' : 'info';
    const debug = !live;
    const radarOptions = Object.assign(
      Config.defaultOptions,
      {
        publishableKey,
        live,
        logLevel,
        debug,
      },
      options,
    );
    Config.setup(radarOptions);

    Logger.info(`initialized with ${live ? 'live' : 'test'} publishableKey.`);
    if (options.debug) {
      Logger.info(`using options: ${JSON.stringify(options)}`);
    }

    return ConfigAPI.getConfig();
```

src/api/context.ts
```diff
@@ -0,0 +1,54 @@
...
+class ContextAPI {
+  public static async getContext(location: Location): Promise<RadarContextResponse> {
+    const options = Config.get();
+
+    // get device location if coordinates not provided
+    if (!location.latitude || !location.longitude) {
+      location = await Navigator.getCurrentPosition();
+    }
+
+    const { latitude, longitude, accuracy } = location;
```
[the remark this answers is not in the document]
kochis: Oh it should be optional

src/index.ts
kochis: No, this should happen in the background (does not need to block the call)

PR #107, opened 2023-07-24
expose trackVerified()

nickpatrick:
Exposes `trackVerified()`, which calls `POST /track` on desktop agent running on `localhost:52516`. Need to add error handling (e.g., case where desktop agent not installed or not running)

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -4,6 +4,7 @@ import Logger from './logger';
 
 import {
   RadarBadRequestError,
+  RadarDesktopAppError,
   RadarForbiddenError,
   RadarNotFoundError,
   RadarPaymentRequiredError,
@@ -131,7 +132,11 @@ class Http {
       }
 
       xhr.onerror = function() {
-        reject(new RadarServerError());
+        if (host && host.includes('localhost')) {
+          reject(new RadarDesktopAppError());
+        } else {
+          reject(new RadarServerError());
+        }
       }
 
       xhr.ontimeout = function() {
--- a/src/http.ts
+++ b/src/http.ts
@@ -132,7 +132,7 @@ class Http {
       }
 
       xhr.onerror = function() {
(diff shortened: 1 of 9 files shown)


tderouin:
I tested this out with Radar Verify, and it worked well.

src/index.ts
```diff
@@ -143,6 +144,14 @@ class Radar {
     }
   }
 
+  public static trackVerified(params: RadarTrackParams = {}) {
+    try {
+      return VerifyAPI.trackVerified(params);
+    } finally {
+      ConfigAPI.getConfig(params); // call with updated permissions
```
kochis, line 151: It doesn't seem like trackVerified is making any calls that would modify permissions

src/http.ts
```diff
@@ -27,10 +27,12 @@ class Http {
     method,
     path,
     data,
+    host = undefined,
   }: {
     method: HttpMethod;
     path: string;
     data?: any;
+    host?: string | undefined;
```
kochis, line 35: I think this can just be `host?: string;`, then you don't need the default argument above.

src/index.ts
nickpatrick: Agreed, will simplify

src/http.ts
```diff
@@ -129,7 +132,11 @@ class Http {
       }
 
       xhr.onerror = function() {
-        reject(new RadarServerError());
+        if (host && host.includes('localhost')) {
```
kochis, line 135: We should include the port here, it's possible that localhost is valid (ie. developing against dashboard locally)

src/index.ts
nickpatrick: Done

src/http.ts
nickpatrick: Done

kochis (approved):

PR #108, opened 2023-07-25
Add attribution to map by default

kochis:
* Enable map attribution


kochis:
Close in favor of https://github.com/radarlabs/radar-sdk-js/pull/109

PR #109, opened 2023-07-25
Add default attribution

kochis:
Default to showing map attribution, using the non-compact style

Screenshot 2023-07-25 at 1 58 16 PM

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -53,7 +53,6 @@ class MapUI {
       { style },
       mapOptions,
     );
-    console.log('PASSED OPTIONS', mapOptions);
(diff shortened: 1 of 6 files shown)


src/ui/map.ts
```diff
@@ -79,16 +82,28 @@ class MapUI {
     link.appendChild(img)
     map.getContainer().appendChild(link);
 
+    // add attribution
+    const attribution = new maplibregl.AttributionControl({ compact: false });
```
corypisano, line 86: can we use the default resizing/hiding attribution?

corypisano (approved):
looks good just confirming the `compact: false` need / managing the resizing

PR #110, opened 2023-07-26
support providing input element for autocomplete ui

kochis:
* Adds support for providing the `<input>` for autocomplete UI

image


kochis:
Yeah, that’s the intention. I’ll update the docs to make it more clear.

PR #119, opened 2023-08-21
add trackVerifiedToken()

nickpatrick:
- exposes `Radar.trackVerifiedToken()`
- adds a new demo page

image

### DIFF
--- a/demo/views/track-verified-token.hbs
+++ b/demo/views/track-verified-token.hbs
@@ -21,9 +21,9 @@
   </div>
 
   <div id="response-details" style="display: none;">
-    <h6 class="mt-2">Token</h6>
+    <h6 class="mt-4">Token</h6>
     <div class="mt-2">
-      <pre><code id="token" class="code-sample language-json"></code><pre>
+      <pre style="white-space: pre-wrap"><code id="token" class="code-sample language-json"></code><pre>
     </div>
   </div>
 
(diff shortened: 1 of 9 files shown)


demo/views/track-verified-token.hbs
```diff
@@ -0,0 +1,94 @@
...
+  </div>
+
+  <div id="response-details" style="display: none;">
+    <h6 class="mt-4">Token</h6>
+    <div class="mt-2">
+      <pre style="white-space: pre-wrap"><code id="token" class="code-sample language-json"></code><pre>
+    </div>
+  </div>
+
+  <div class="mt-4">
```
kochis, line 30: We can probably omit the code samples on these. These should probably live in the docs (I only added them as a temporary solution until we go the UI kits in the docs).

src/api/verify.ts
```diff
@@ -49,27 +50,39 @@ class VerifyAPI {
...
+      const trackTokenRes = {
+        token,
+      } as RadarTrackTokenResponse;
+  
+      if (options.debug) {
+        trackTokenRes.response = response;
+      }
+  
+      return trackTokenRes;
+    } else {
```
kochis, line 73: Technically don't need this `else` since it returns above.

kochis (dismissed):

demo/views/track-verified-token.hbs
nickpatrick: I've actually found it helpful to be able to show the code sample when using this page in demos (which I'm doing a lot), but we can always remove later

src/api/verify.ts
nickpatrick: Updated

PR #121, opened 2023-08-28
trackVerified() behavior by platform

nickpatrick:


src/api/track.ts
```diff
@@ -9,10 +9,10 @@ import Storage from '../storage';
 
 import TripsAPI from './trips';
 
-import type { RadarTrackParams, RadarTrackResponse } from '../types';
+import type { RadarTrackParams, RadarTrackResponse, RadarTrackTokenResponse } from '../types';
 
 class TrackAPI {
-  static async trackOnce(params: RadarTrackParams) {
+  static async trackOnce(params: RadarTrackParams, verified: boolean = false, encrypted: boolean = false, host: string | undefined = undefined) {
```
kochis, line 15: Lets just roll these into `RadarTrackParams`

src/api/verify.ts
```diff
@@ -9,7 +9,7 @@ import Storage from '../storage';
 import type { RadarTrackParams, RadarTrackResponse, RadarTrackTokenResponse } from '../types';
 
 class VerifyAPI {
-  static async trackVerified(params: RadarTrackParams, encrypted: Boolean = false) {
+  static async trackVerified(params: RadarTrackParams, encrypted: Boolean = false, host: string = 'http://localhost:52516') {
```
kochis, line 12: Same as above, lets just make `encrypted` part of track params

src/device.ts
```diff
@@ -11,6 +12,23 @@ const generateUUID = (): string => {
 };
 
 class Device {
+  static getPlatform(): string {
+    let userAgent = navigator.userAgent;
```
kochis, line 16: Any idea how reliable this is? UserAgent parsing can have a lot of edge cases

src/api/track.ts
kochis: Or add another object that is like `verifiedObjects` or something
```ts
static async trackOnce(params: RadarTrackParams, verifiedOptions?: RadarVerifiedOptions) {
```

nickpatrick: I thought about adding it to `RadarTrackParams`, but we don't want to allow developers to pass it explicitly. More an internal implementation detail. Whether we pass `encrypted = true` (which we will prob rename or remove, btw) is implicit in whether the developer is calling `trackVerified()` or `trackVerifiedToken()

src/index.ts
```diff
@@ -136,20 +139,42 @@ class Radar {
...
+      return VerifyAPI.trackVerified(params, false, 'https://radar-verify.com:52516');
+    } else if (platform === 'DESKTOP_WINDOWS') { // use Windows app
+      return VerifyAPI.trackVerified(params, false, 'http://localhost:52516');
+    } else { // bypass desktop app
+      try {
+        return TrackAPI.trackOnce(params, true, false, 'https://api-verified.radar.io');
+      } finally {
+        ConfigAPI.getConfig(params); // call with updated permissions
+      }
+    }
```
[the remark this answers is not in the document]
nickpatrick: Can update

src/device.ts
nickpatrick: I think it's the best option for this use case. There's also `navigator.platform`, but that's deprecated. We could also let the developer decide (e.g., `Radar.trackVerified({ useDesktopApp: boolean = false })`), but that just shifts the problem to the developer

src/api/track.ts
kochis: We should still wrap it in an optional object. Otherwise it's hard to tell what the arguments are at the callsite:
```ts
TrackAPI.trackOnce(params, true, false, 'https://api-verified.radar.io');
```
As opposed to:
```ts
TrackAPI.trackOnce(params, {
  verified: true, 
  encrypted: false, 
  host: 'https://api-verified.radar.io',
});
```

nickpatrick:
Closing in favor of simpler #126

Issue #124, opened 2023-09-12
Incorrect Typescript interface


cuginoAle:
Hi, 
while experimenting with Radar Js SDK I noticed the following:

<img width="872" alt="image" src="https://github.com/radarlabs/radar-sdk-js/assets/1298616/91946e3e-3a1b-4f65-8b5d-751c0d0d0951">

It looks like the `Radar.autocomplete` promise returns addresses as a single `RadarAutocompleteAddress` rather than an array:

<img width="626" alt="image" src="https://github.com/radarlabs/radar-sdk-js/assets/1298616/19fd10db-9645-412d-8def-8f6f27056937">

Should it return `RadarAutocompleteAddress[]` instead?

kochis:
Hi @cuginoAle,

Thanks for catching, that does look to be a bug. I added a PR with the updating typings here: https://github.com/radarlabs/radar-sdk-js/pull/125

Should be able to get that fix in soon.

kochis:
Fixed in [v4.1.7](https://github.com/radarlabs/radar-sdk-js/releases/tag/v4.1.7)

PR #125, opened 2023-09-12
Ensure array of addresses is defined in RadarAutocompleteResponse

kochis:
Fixes https://github.com/radarlabs/radar-sdk-js/issues/124

* `addressess` in `RadarAutocompleteResponse` should be an array of addresses.


jaspk06 (approved):
👍

PR #126, opened 2023-09-19
Track verified updates

nickpatrick:


src/http.ts
```diff
@@ -96,6 +98,15 @@ class Http {
           return reject(new RadarServerError(response));
         }
 
+        const error = response?.meta?.error;
+        if (error === 'ERROR_PERMISSIONS') {
+          return reject(new RadarLocationPermissionsError('Location permissions not granted.'));
+        } else if (error === 'ERROR_LOCATION') {
+          return reject(new RadarLocationError('Could not determine location.'));
+        } else if (error === 'ERROR_NETWORK') {
+          return reject(new RadarTimeoutError());
```
nickpatrick, line 107: @kochis We might want to update this to `RadarNetworkError` to match other SDKs (and be inclusive of network errors other than timeouts) in future web SDK versions, but since it's a breaking change we can punt for now. 

Desktop app will return `{ meta: { error: 'ERROR_PERMISSIONS' | 'ERROR_LOCATION' | 'ERROR_NETWORK' } }` from `localhost` `GET /verify` calls

kochis: Yeah, I forget why it's a timeout erorr, but makes sense to be more generic.

kochis (approved):

PR #127, opened 2023-09-20
hideResultsOnBlur and setOptions

jaspk06:
Added `hideResultsOnBlur` option so that the results list persist on click away.

Also, added `setOptions` function so that the autocomplete element does not have to be completely rerendered everytime an option changes.

These are needed for internal usage

### QA
https://www.loom.com/share/2084a71a836041ebbed533a88fe7a27c

### DIFF
--- a/src/ui/autocomplete.ts
+++ b/src/ui/autocomplete.ts
@@ -492,10 +492,33 @@ class AutocompleteUI {
     // update height
     setHeight(this.resultsList, this.config);
 
-    // update marker color
-    const marker = this.resultsList.getElementsByClassName(CLASSNAMES.RESULTS_MARKER)[0];
-    if (marker) {
+    // update showMarkers
+    if (this.config.showMarkers === true) {
+      const marker = document.createElement('img');
+      marker.classList.add(CLASSNAMES.RESULTS_MARKER);
       marker.setAttribute('src', getMarkerIcon(this.config.markerColor));
+      const resultItems = this.resultsList.getElementsByTagName('li');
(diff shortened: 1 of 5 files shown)


src/ui/autocomplete.ts
```diff
@@ -458,6 +461,98 @@ class AutocompleteUI {
...
+  }
+
+  public setLimit(limit: number) {
+    this.config.limit = limit;
+    return this;
+  }
+
+  public setShowMarkers(showMarkers: boolean) {
+    this.config.showMarkers = showMarkers;
+    if (showMarkers === true) {
```
kochis, line 516: We can probably just assume a truthy value here `if (showMarkers)`

src/ui/autocomplete.ts
```diff
@@ -50,8 +51,9 @@ const setWidth = (input: HTMLElement, options: RadarAutocompleteUIOptions) => {
     return;
   }
 
-  // if not responsive, set fixed width
+  // if not responsive, set fixed width and unset maxWidth
   input.style.width = formatCSSValue(options.width || DEFAULT_WIDTH);
+  input.style.maxWidth = 'none';
```
kochis, line 56: might be better to just set as an empty string (or call `removeProperty`): https://stackoverflow.com/questions/2027935/how-to-remove-css-property-using-javascript

kochis (approved):

src/ui/autocomplete.ts
jaspk06: ah that sounds like a better idea. will do that

kochis (approved):

PR #136, opened 2023-11-06
Adding radar-core cdn version.

yortyrh:
**Problem:** The javascript file has 762K.
**Solution:** Create a version which contains only the API and not the UI.
**Result:** File with only 26K.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -50,6 +50,31 @@ export default [
 
   // IIFE (browser bundles, written to /cdn)
   {
+    input: 'src/core.ts',
+    output: [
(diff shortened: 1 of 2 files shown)


yortyrh:
What do you think about this one @jaspk06 ? Any suggestion?

kochis:
Merged in https://github.com/radarlabs/radar-sdk-js/pull/171

You can find the released builds here: 
* (latest) https://js.radar.com/v4/radar-core.min.js
* (pinned) https://js.radar.com/v4.4.1/radar-core.min.js

PR #139, opened 2023-12-07
[WIP] MAPS-694 clean up groups from sdk functions 

j949wang:


kochis:
There probably shouldn't be package-lock changes for this.

PR #141, opened 2023-12-12
MAPS-748 autocomplete sessions into javascript sdk

boonboonsiri:


src/ui/autocomplete.ts
```diff
@@ -511,6 +548,10 @@ class AutocompleteUI {
     return this;
   }
 
+  public generateUUID = (): string => {
+    return crypto.randomUUID();
```
boonboonsiri, line 552: I wanted to use crypto.randomUUID() but not sure if this is the case but maybe:

https://github.com/parcel-bundler/parcel/issues/8820

src/ui/autocomplete.ts
```diff
@@ -442,6 +457,28 @@ class AutocompleteUI {
       onSelection(result);
     }
 
+    // Radar Autocomplete Session
+    if (this.session) {
+      const userId = Storage.getItem(Storage.USER_ID) || undefined;
```
boonboonsiri, line 462: Right now I'm still including these even though right now we don't use them, but in case we wanted to store them in the future so we don't have to update the sdk twice

PR #143, opened 2023-12-26
Adding an alt text to the radar map logo

LucasFSCAR:
Simply putting and alt to the map logo to improve SEO.


nickpatrick:
Thanks! @kochis I might suggest alt text of `Radar Maps Platform`

PR #147, opened 2024-01-16
JWT signing and incognito mode detection

nickpatrick:


package.json
```diff
@@ -49,7 +49,9 @@
     "typescript": "^5.0.4"
   },
   "dependencies": {
-    "@types/geojson": "^7946.0.10"
+    "@types/geojson": "^7946.0.10",
+    "detectincognitojs": "^1.3.0",
```
kochis, line 53: This package is just a single file: https://github.com/Joe12387/detectIncognito/blob/main/src/detectIncognito.ts

We could probably pull it into a local `utils/` file.

src/api/track.ts
```diff
@@ -1,3 +1,7 @@
+import { detectIncognito } from 'detectincognitojs';
+// @ts-ignore
+import { KJUR } from 'jsrsasign';
```
kochis, line 3: I think we can just import `{ jws }` instead of `KJUR`, which is the GitHub author's name... https://github.com/kjur/jsrsasign/blob/master/npm/lib/jsrsasign.js#L341

Also, this is a pretty massive library to be including just for this use case. We should see if there's a more appropriate package, or just pull in the pieces we need (this code probably doesn't change very often).

src/api/track.ts
```diff
@@ -85,11 +88,64 @@ class TrackAPI {
...
+      const header = JSON.stringify({
+        alg: 'HS256',
+        typ: 'JWT',
+      });
+      const payload = JSON.stringify({
+        payload: JSON.stringify({
+          ...body,
+          incognito,
+        }),
+      });
```
kochis, line 127: We probably only need to call `stringify` once:
```js
const data = JSON.stringify({
  payload: {
    ...body,
    incognito,
  },
});
```

src/http.ts
```diff
@@ -143,7 +153,7 @@ class Http {
       }
 
       xhr.onerror = function() {
-        if (host && host === 'https://radar-verify.com:52516') {
+        if (host && (host === 'http://localhost:52516' || host === 'https://radar-verify.com:52516')) {
```
kochis, line 156: We have like 3 different hard-coded URLs now, we should probably have them in a config or constants file somewhere:
```
https://api-verified.radar.io
https://radar-verify.com:52516
http://localhost:52516
```

src/types.ts
```diff
@@ -73,6 +73,7 @@ export interface RadarTrackParams {
   metadata?: RadarMetadata;
   tripOptions?: RadarTripOptions;
   desiredAccuracy?: 'high' | 'medium' | 'low';
+  fraud?: boolean;
```
kochis, line 76: Is this what's is called in mobile SDKs? This boolean acts as a flag, so a verb might be more appropriate (`useFraud`, `checkFraud`, `enableFraud` etc...)

PR #149, opened 2024-01-21
latency calcs with web sockets

nickpatrick:


src/api/track.ts
```diff
@@ -115,18 +109,35 @@ class TrackAPI {
...
+            method: 'GET',
+            path: 'ping',
+          }),
+          ping(`wss://${pingHost}`),
+        ]);
+        const { scl }: any = sclRes;
+        sclVal = scl;
+        cslVal = csl;
+      } catch (err) {
+
```
kochis, line 127: Could do a `Logger.warn` here

src/api/track.ts
```diff
@@ -139,6 +150,17 @@ class TrackAPI {
...
+      if (response && response.user) {
+        if (!response.user.metadata) {
+          response.user.metadata = {};
+        }
+
+        response.user.metadata['radar:debug'] = {
+          sclVal,
+          cslVal,
+        };
```
kochis, line 162: Could wrap this in `options.debug` if its not applicable to all users 

(gets logged when SDK is initialized with `debug: true`)

src/api/track.ts
```diff
@@ -11,6 +7,8 @@ import Navigator from '../navigator';
 import Session from '../session';
 import Storage from '../storage';
 import TripsAPI from './trips';
+import { signJWT } from '../util/jwt'
+import { ping } from '../util/ping'
```
kochis, line 11: Could maybe call this file `util/network` or just `util/net`

src/util/ping.ts
```diff
@@ -0,0 +1,52 @@
...
+            const median = latencies.sort((a, b) => a - b)[1];
+            socket.close();
+            resolve(median);
+          }
+        }
+      };
+    }
+
+    const timeout = () => {
+      console.error('Socket timeout');
```
kochis, line 33: Should make these `Logger.error` (or `Logger.warn`)

src/api/track.ts
nickpatrick: Just added this to print on ping.radar-verify.com (not console). Can remove later. If there's a better way to stash than updating `metadata`, lmk

src/api/track.ts
nickpatrick: Can update

src/util/ping.ts
nickpatrick: Good call

src/api/track.ts
nickpatrick: Rather be somewhat opaque at this step, and we're logging errors server-side

kochis: Maybe add a comment `// do nothing`, the empty block just looks sorta odd (though obvious what its doing)

nickpatrick: Done

PR #152, opened 2024-03-27
Custom styles

jaspk06:
NOTE:
This is the FIRST PR that should be reviewed as a part of the maps studio + web sdk improvement projects. TBH a lot of these changes are kinda irrelevant since they're gonna be superseded by this [refactor PR](https://github.com/radarlabs/radar-sdk-js/pull/153). More emphasis should be put on that following PR

this PR covers the following changes:
- supports custom styles (but not custom markers)
- uses a Maplibre control component for the radar logo

the next PR to review is: https://github.com/radarlabs/radar-sdk-js/pull/153

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -31,12 +31,14 @@ class Http {
     path,
     data,
     host,
-    headers,
+    versioned = true,
+    headers = {},
   }: {
     method: HttpMethod;
     path: string;
     data?: any;
     host?: string;
+    versioned?: boolean;
     headers?: Record<string, string>;
   }) {
     return new Promise<HttpResponse>((resolve, reject) => {
@@ -53,6 +55,10 @@ class Http {
       const urlHost = host || options.host;
       const version = options.version;
       let url = `${urlHost}/${version}/${path}`;
+
+      if (!versioned) {
+        url = `${urlHost}/${path}`;
+      }
 
       // remove undefined values from request data
       let body: any = {};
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -15,7 +15,7 @@ import TrackAPI from './api/track';
 import TripsAPI from './api/trips';
 import VerifyAPI from './api/verify';
 
-import MapUI from './ui/map';
+import MapUI from './ui/mapFactory';
```
kochis, line 18: Lets just keep this file as `/ui/map`

src/ui/mapFactory.ts
```diff
@@ -0,0 +1,26 @@
+import maplibregl from 'maplibre-gl';
+
+import RadarMap from './map';
+import RadarMarker from './marker';
+
+import type { RadarMapOptions, RadarMarkerOptions } from '../types';
+
+class MapUI {
```
kochis, line 8: Lets keep this file as `/src/ui/map` as the main entrypoint to the maps stuff.

This can import the other files, and their file names can match the class names
```js
import RadarMap from './RadarMap';
import RadarMarker from './RadarMarker';
```

src/ui/marker.ts
```diff
@@ -0,0 +1,100 @@
+import maplibregl, { MapMouseEvent } from 'maplibre-gl';
+
+import RadarMap from './map';
```
kochis, line 3: What does this file need the map reference for, just for typing? Also looks like this is a circular dependency.

jaspk06: yeah it's for typing so technically it's okay*. I can explicitly import the type instead

jaspk06: refactor now lives in this pr: https://github.com/radarlabs/radar-sdk-js/pull/153

src/http.ts
```diff
@@ -31,12 +31,14 @@ class Http {
...
     host,
-    headers,
+    versioned = true,
+    headers = {},
   }: {
     method: HttpMethod;
     path: string;
     data?: any;
     host?: string;
+    versioned?: boolean;
```
kochis, line 41: I think instead of `versioned` as a boolean, we can maybe accept `version` as an optional string here. 

In this case, the version is actually `maps`.

So below, this can take precedence if it's set in the request
```js
const urlHost = host || options.host;
const urlVersion = version || options.version;
let url = `${urlHost}/${version}/${path}`;
```

src/ui/map.ts
```diff
@@ -28,22 +30,49 @@ const defaultMarkerOptions: Partial<maplibregl.MarkerOptions> = {
...
+    rotationAlignment: marker.getRotationAlignment(),
+    pitchAlignment: marker.getPitchAlignment()
+  }
+
+  return markerOptions;
+}
+
+/** Check if style is a Radar style or a custom style */
+const isRadarStyle = (style: string) => {
+  return RADAR_STYLES.includes(style) || uuidRegex.test(style)
```
kochis, line 58: I'm not so sure we need to validate the UUID here. We should maybe just consider any style to be a Radar style.

The other thing we could do, is prefix the UUIDs with `radar://`. Probably a larger discussion.

src/ui/map.ts
```diff
@@ -121,15 +138,59 @@ class MapUI {
...
+
+    const onStyleLoad = async () => {
+      MapUI.customMarkerRawSvg = undefined;
+      const style = map.getStyle();
+
+      const customMarkers = (style.metadata as any)?.['radar:customMarkers'];
+      if (Array.isArray(customMarkers) && customMarkers.length > 0) {
+        const customMarker = customMarkers[0]; // only support one custom marker for now
+        try {
+          const markerRawSvg = await Http.request({
```
kochis, line 150: Rather than fetching the marker content, I wonder if we can just create an image element. Maybe there's some drawbacks to that.
```html
<img src="https://api.radar.io/maps/markers/${marker.id}?publishableKey=${publishableKey}" />
```

src/ui/map.ts
```diff
@@ -121,15 +138,59 @@ class MapUI {
...
+              'Content-Type': 'image/svg+xml',
+            },
+          });
+          MapUI.customMarkerRawSvg = markerRawSvg.data;
+        } catch (err) {
+          Logger.warn(`Error getting custom marker: ${customMarker.id} - using default marker.`);
+        }
+      }
+
+      // set markers if necessary
```
kochis, line 164: This is to handle dynamic style changes? One thing to keep in mind, we should only do this if it is already using a default marker.

src/ui/map.ts
```diff
@@ -121,15 +144,59 @@ class MapUI {
         }
       }
     };
+
+    const onStyleLoad = async () => {
```
jaspk06, line 148: NOTE: this `onStyleLoad` hook is to handle custom markers that uses an old pattern that was previously discussed. We've since decided that rather than handling the default marker for them (i.e. defaulting to the custom marker), we're just gonna have them pass in the custom marker id. This change is covered in this follow up PR: https://github.com/radarlabs/radar-sdk-js/pull/153

kochis:
Closing in favor of https://github.com/radarlabs/radar-sdk-js/pull/153

PR #153, opened 2024-04-10
Radar map UI kit refactor

jaspk06:
uses an inheritance pattern instead of wrapping the `maplibre-gl Map` in helper functions. This allows us to more closely develop features on top of the `maplibre-gl Map` class instance

### DIFF
--- a/src/ui/RadarMarker.ts
+++ b/src/ui/RadarMarker.ts
@@ -1,0 +1,86 @@
+import maplibregl from 'maplibre-gl';
+
+import type RadarMap from './RadarMap';
+
+import type { RadarMarkerImage, RadarMarkerOptions } from '../types';
+
+const defaultMarkerOptions: maplibregl.MarkerOptions = {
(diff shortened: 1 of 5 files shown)


src/ui/RadarMarker.ts
```diff
@@ -0,0 +1,126 @@
...
+            if (res.status === 200) {
+              res.blob()
+                .then(onSuccess)
+                .catch(onError);
+            } else {
+              onError(new Error(res.statusText));
+            }
+          })
+          .catch(onError)
+      }
```
jaspk06, line 88: not the prettiest way to handle this but we can't use `async await` in a constructor. if `markerOptions.image.url` exists, technically we can just do something like `img.src = markerOptions.image.url` but writing this way allows the marker to fallback to the default if there is an error with the provided url.

kochis (approved):
Looks good. Lets start releasing the map changes as 4.3.0-beta.X

PR #154, opened 2024-04-10
Marker click event

jaspk06:
NOTE: https://github.com/radarlabs/radar-sdk-js/pull/153 needs to be reviewed first

- adds a 'click' event to `RadarMarker`. This is useful for attaching event listeners to the marker

### DIFF
--- a/src/api/track.ts
+++ b/src/api/track.ts
@@ -137,14 +137,14 @@ class TrackAPI {
         }),
       };
       
-      const token = await signJWT(payload, dk);
+      const reqToken = await signJWT(payload, dk);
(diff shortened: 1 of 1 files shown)


src/ui/RadarMarker.ts
```diff
@@ -64,6 +64,20 @@ class RadarMarker extends maplibregl.Marker {
     return super.remove();
   }
 
+  _onMapClick(e: maplibregl.MapMouseEvent): void {
```
kochis, line 67: Is this an inherited function / how is this called?

PR #156, opened 2024-04-10
`RadarMap.fitToMarkers` class method

jaspk06:
NOTE: https://github.com/radarlabs/radar-sdk-js/pull/153 needs to be reviewed first

single line function to adjust the bounds of a `RadarMap` to fit to all `RadarMarker` added to the map.


src/ui/RadarMap.ts
```diff
@@ -155,6 +155,15 @@ class RadarMap extends maplibregl.Map {
     }
     this.on('styledata', onStyleLoad);
   }
+
+  fitToMarkers(fitBoundsOptions: maplibregl.FitBoundsOptions = { padding: 50 }, overrideMarkers?: RadarMarker[]) {
+    const bounds = new maplibregl.LngLatBounds();
+    const markers = overrideMarkers || this._markers;
+    markers.forEach(marker => {
```
kochis, line 162: We should probably add a lint rule for this, but there should be parens around the arguments
```
markers.forEach((markers) => {
```

src/ui/RadarMap.ts
```diff
@@ -155,6 +155,15 @@ class RadarMap extends maplibregl.Map {
...
+      bounds.extend(marker.getLngLat());
+    });
+    this.fitBounds(bounds, fitBoundsOptions);
```
kochis, line 165: What does this do if markers is empty?

jaspk06: added a no-op if markers is empty. Yeah you can pass in animation options into `fitBoundsOptions` https://docs.mapbox.com/mapbox-gl-js/api/map/#map#fitbounds

jaspk06: just do `fitMarkers({animate: false})` to disable animations. it uses `map.flyTo` by default

src/ui/RadarMap.ts
```diff
@@ -135,6 +135,20 @@ class RadarMap extends maplibregl.Map {
   getMarkers(): RadarMarker[] {
     return this._markers;
   }
+
+  fitToMarkers(fitBoundsOptions: maplibregl.FitBoundsOptions = { padding: 50 }, overrideMarkers?: RadarMarker[]) {
+    const bounds = new maplibregl.LngLatBounds();
```
kochis, line 140: nit: move this down to just before the forEach. No need to initialize if we're going to return early.

kochis (approved):

PR #159, opened 2024-05-06
Support new expiresAt/expiresIn/token/passed fields with trackOnce({ fraud: true }) and trackVerified()

nickpatrick:
- When calling `trackOnce({ fraud: true })`, return new `token/expiresAt/expiresIn/passed` fields. @tderouin right now we're calculating `expiresIn` and `passed` on the client, but cleaner to calculate them on the server. Could be a fast follow
- Consolidates `trackVerified()` and `trackVerifiedToken()`. Now just a single function which returns all of `user/events/location/token/expiresAt/expiresIn/passed`. No desktop app changes required
- Renames error codes to match mobile SDKs for consistency. `RadarTimeoutError` --> `RadarNetworkError`, `RadarLocationPermissionsError` --> `RadarPermissionsError`. Also renames `RadarDesktopAppError` to `RadarVerifyAppError` given the addition of the Radar Verify mobile apps
- Fixes long-standing issue where we return `RadarLocationPermissionsError` for any location error, including location timeouts. Now we only return `RadarPermissionsError` for true permissions errors (code 1) and `RadarLocationError` for location timeouts or unable to determine location (codes 2 and 3)


src/types.ts
```diff
@@ -212,8 +213,11 @@ export interface RadarTrackResponse extends RadarResponse {
   events?: RadarEvent[];
 }
 
-export interface RadarTrackTokenResponse extends RadarResponse {
+export interface RadarTrackVerifiedResponse extends RadarTrackResponse {
```
nickpatrick, line 216: Could also be a separate interface instead of extending `RadarTrackResponse`

src/errors.ts
```diff
@@ -23,19 +23,19 @@ export class RadarLocationError extends RadarError {
   }
 }
 
-export class RadarLocationPermissionsError extends RadarError {
+export class RadarPermissionsError extends RadarError {
   constructor(message: string) {
     super(message);
-    this.name = 'RadarLocationPermissionsError';
+    this.name = 'RadarPermissionsError';
```
nickpatrick, line 29: To match mobile SDKs

src/errors.ts
```diff
@@ -23,19 +23,19 @@ export class RadarLocationError extends RadarError {
...
 }
 
-export class RadarDesktopAppError extends RadarError {
+export class RadarVerifyAppError extends RadarError {
   constructor() {
-    super('Desktop app not running.');
-    this.name = 'RadarDesktopAppError';
-    this.status = 'ERROR_DESKTOP_APP';
+    super('Radar Verify app not running.');
+    this.name = 'RadarVerifyAppError';
```
nickpatrick, line 37: To account for Radar Verify desktop AND mobile apps

src/errors.ts
```diff
@@ -131,11 +131,11 @@ export class RadarServerError extends RadarError {
   }
 }
 
-export class RadarTimeoutError extends RadarError {
+export class RadarNetworkError extends RadarError {
   constructor() {
     super('Request timed out.');
-    this.name = 'RadarTimeoutError';
-    this.status = 'ERROR_TIMED_OUT';
+    this.name = 'RadarNetworkError';
```
nickpatrick, line 137: To match mobile SDKs

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "radar-sdk-js",
-  "version": "4.1.18",
+  "version": "4.2.0-beta.0",
```
kochis, line 3: Are we shipping these changes as `4.2`?

src/api/track.ts
```diff
@@ -160,15 +160,41 @@ class TrackAPI {
...
         };
+
+        let { user, events, token, expiresAt } = response;
+        const location = { latitude, longitude, accuracy };
+        let passed;
+        let expiresIn;
+        if (expiresAt) {
+          expiresAt = new Date(expiresAt);
+          passed = user?.fraud?.passed && user?.country?.passed && user?.state?.passed;
+          expiresIn = expiresAt ? (expiresAt.getTime() - new Date().getTime()) / 1000 : null;
```
kochis, line 171: nit: It looks like `expiresAt` will also be present here.

nit: You can also do `Date.now()` instead of `new Date().getTime()`

kochis (dismissed):

package.json
nickpatrick: Removing `trackVerifiedToken()` and renaming the error codes are breaking changes, so should prob be `4.2.x` instead of `4.1.x`

src/api/track.ts
nickpatrick: Updated. @tderouin if we can expose `expiresIn` (in seconds, `expiresAt - now`) in the API response, I can also just remove this, along with `passed`

src/api/track.ts
```diff
@@ -160,15 +160,41 @@ class TrackAPI {
...
         };
+
+        let { user, events, token, expiresAt } = response;
+        const location = { latitude, longitude, accuracy };
+        let passed;
+        let expiresIn;
+        if (expiresAt) {
+          expiresAt = new Date(expiresAt);
+          passed = user?.fraud?.passed && user?.country?.passed && user?.state?.passed;
+          expiresIn = (expiresAt.getTime() - Date.now()) / 1000;
```
david-goodfellow, line 171: should the name be `expiresInSeconds` to be explicit? No strong opinion here because jwt uses `expiresIn` as well and just has the docs saying it is in seconds

src/index.ts
```diff
@@ -148,10 +148,6 @@ class Radar {
     return VerifyAPI.trackVerified(params);
   }
 
-  public static trackVerifiedToken(params: RadarTrackParams = {}) {
```
kochis, line 151: If we don't want this to be a breaking change, could we potentially pass the `params` here to track width `{ fraud: true }`?

Could also include a deprecation notice:
```js
Logger.warn('DEPRECATED - trackVerifiedToken will be removed in a future release.');
```

kochis (approved):

src/index.ts
nickpatrick: Nobody uses this in prod yet (except for our demo page), so I think we're fine to skip this

src/api/track.ts
nickpatrick: We don't really include units in the field name anywhere else in our APIs, so prefer to keep it `expiresIn` (if I could go back in a time machine to 2017 I might have used this pattern, but too late now, rather optimize for consistency and brevity)

src/api/track.ts
tderouin: https://github.com/radarlabs/server/pull/5823/commits/a8fd054e1f02f5e2b9a1c44f9cde6f44927db03c

PR #160, opened 2024-05-09
v4.3

kochis:
Features:
* Support for custom map styles
* Language support for maps
* Marker enhancements:
  * Custom marker image support
  * Marker `.on('click')` handler
  * `map.fitToMarkers()` function

Enhancements:
* Support for MapLibre 4.x
* Build output from es5 -> es6
* Send `X-Sdk-Version` for Maps requests

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -31,13 +31,17 @@ class Http {
     path,
     data,
     host,
-    headers,
+    version,
+    headers = {},
+    responseType,
   }: {
     method: HttpMethod;
     path: string;
     data?: any;
     host?: string;
+    version?: string;
     headers?: Record<string, string>;
+    responseType?: XMLHttpRequestResponseType;
   }) {
     return new Promise<HttpResponse>((resolve, reject) => {
       const options = Config.get();
@@ -51,8 +55,8 @@ class Http {
 
       // setup request URL
       const urlHost = host || options.host;
-      const version = options.version;
-      let url = `${urlHost}/${version}/${path}`;
+      const urlVersion = version || options.version;
+      let url = `${urlHost}/${urlVersion}/${path}`;
 
       // remove undefined values from request data
       let body: any = {};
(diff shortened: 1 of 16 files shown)


src/ui/RadarMarker.ts
```diff
@@ -0,0 +1,170 @@
...
+const createImageElement = (options: ImageOptions) => {
+  const element = document.createElement('img');
+  element.src = options.url!;
+
+  if (options.width) {
+    if (typeof options.width === 'number') {
+      element.width = options.width;
+    } else {
+      element.style.width = options.width;
+    }
```
jaspk06, line 26: if we're handling both `number` and `string` should we change

```ts
export interface RadarMarkerOptions extends maplibregl.MarkerOptions {
  /**
   * @deprecated Use popup.text
   */
  text?: string;

  /**
   * @deprecated Use popup.html
   */
  html?: string;

  // marker configs
  url?: string;
  marker?: string;
  width?: number | string; // accept string now
  height?: number | string; // accept string now

  // popup configs
  popup?: RadarMarkerPopupOptions;
}
```

src/http.ts
```diff
@@ -78,32 +82,39 @@ class Http {
...
 
       xhr.onload = () => {
         let response: any;
         try {
-          response = JSON.parse(xhr.response);
+          if (xhr.responseType === 'blob') {
+            response = { code: xhr.status, data: xhr.response };
+          } else {
+            response = JSON.parse(xhr.response);
+          }
```
jaspk06, line 117: so ideally this is:
```suggestion
          if (xhr.responseType === 'json') {
            response = JSON.parse(xhr.response);
          } else {
            response = { code: xhr.status, data: xhr.response };
          }
```

but when i wrote this initially i wanted to minimize logic changes so we don't introduce new sweeping bugs

jaspk06 (approved):
lgtm 👍

PR #161, opened 2024-05-10
Map initalization `language` param

jaspk06:
Adds the map option `language` to configure map language. This only works for Radar map styles as it searchs and replaces for the expression `["get", "radar:name_client_language"]`.

Also sends up the header `X-Radar-Device-Language` with every style request which grabs from `navigator.language`. This is so that the server can figure out how to transform the style appropriately given the browser language.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -460,7 +460,7 @@ export interface RadarSearchGeofencesResponse extends RadarResponse {
 }
 
 export interface RadarMapOptions extends Omit<maplibregl.MapOptions, 'transformRequest'> {
-  locale?: string;
+  language?: string;
 }
 
 export interface RadarMarkerImage {
--- a/src/types.ts
+++ b/src/types.ts
@@ -460,7 +460,7 @@ export interface RadarSearchGeofencesResponse extends RadarResponse {
 }
 
 export interface RadarMapOptions extends Omit<maplibregl.MapOptions, 'transformRequest'> {
-  container: string | HTMLElement;
+  locale?: string;
 }
 
 export interface RadarMarkerImage {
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -460,7 +460,7 @@ export interface RadarSearchGeofencesResponse extends RadarResponse {
 }
 
 export interface RadarMapOptions extends Omit<maplibregl.MapOptions, 'transformRequest'> {
-  container: string | HTMLElement;
+  fallbackLanguage?: string;
```
jaspk06, line 463: `container` was already defined in the `maplibregl.MapOptions` type

src/ui/RadarMap.ts
```diff
@@ -69,22 +72,24 @@ class RadarMap extends maplibregl.Map {
...
       if (resourceType === 'Style' && isRadarStyle(url)) {
-        url = createStyleURL(config, url);
+        url = createStyleURL(config, { ...maplibreOptions, style: url });
       }
 
       let headers = {
         'Authorization': config.publishableKey,
         'X-Radar-Device-Type': 'Web',
         'X-Radar-SDK-Version': SDK_VERSION,
+        'X-Radar-Device-Language': navigator.language,
```
jaspk06, line 92: there's also `navigator.languages` which returns an array of the browser's selected languages.  we can use those to order fallback language priority. e.g. `[fr-CA, en-CA]` (candian french fallback to canadian english)

kochis: It might make sense to move this into `sr/navigator` for consistently, and maybe handle the case where navigator is not available.

jaspk06: ah right forgot that existed

jaspk06: done

jaspk06: just returns `undefined` if navigator isn't available

kochis (approved):

PR #162, opened 2024-05-10
fix return, bump version

nickpatrick:


src/api/track.ts
```diff
@@ -160,33 +160,33 @@ class TrackAPI {
           sclVal,
           cslVal,
         };
+      }
```
nickpatrick, line 163: This is really the key change. Ensures the `return` is outside of `if (options.debug ...)`

kochis (dismissed):

PR #168, opened 2024-06-03
Update marker params

kochis:


src/types.ts
```diff
@@ -463,22 +463,30 @@ export interface RadarMapOptions extends Omit<maplibregl.MapOptions, 'transformR
...
-  url?: string;
-  radarMarkerId?: string;
-  width?: number;
-  height?: number;
-}
-
 export interface RadarMarkerPopupOptions extends maplibregl.PopupOptions {
   text?: string;
   html?: string;
+  element?: HTMLElement;
```
jaspk06, line 469: don't think this is being handled anywhere

kochis: Updated: https://github.com/radarlabs/radar-sdk-js/pull/168/files#diff-e53267c1a9488c80008ee9e8ac8ad555734d062f4553b72ba21f7e76323a104fR125

jaspk06 (approved):

PR #170, opened 2024-06-13
Fix autocomplete keyboard control on Windows/Edge

kochis:
The main fix is to change from `event.code` to `event.key` for handling keyboard events, which was breaking on Windows/Edge.

Other enhancements are additional aria-roles for accessibility.

### DIFF
--- a/src/ui/autocomplete.ts
+++ b/src/ui/autocomplete.ts
@@ -16,10 +16,6 @@ const CLASSNAMES = {
   NO_RESULTS: 'radar-no-results',
 };
 
-const ARIA = {
-  EXPANDED: 'aria-expanded',
-};
-
 const defaultAutocompleteOptions: RadarAutocompleteUIOptions = {
   container: 'autocomplete',
   debounceMS: 200, // Debounce time in milliseconds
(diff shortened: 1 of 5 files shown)


src/ui/autocomplete.ts
```diff
@@ -275,6 +283,8 @@ class AutocompleteUI {
     results.forEach((result, index) => {
       const li = document.createElement('li');
       li.classList.add(CLASSNAMES.RESULTS_ITEM);
+      li.setAttribute('role', 'option');
+      li.setAttribute('id', `item-${index}`);
```
jaspk06, line 287: fyi we use the same `class` value for both `class` and `id` attributes for the `input` element but different values for this `li`element. Should this be 
```suggestion
      li.setAttribute('id', `${CLASSNAMES.RESULTS_ITEM}-${index}`);
```
to make it more consistent?

src/ui/autocomplete.ts
```diff
@@ -380,43 +391,41 @@ class AutocompleteUI {
     // add class name to newly highlighted item
     resultItems[index].classList.add(CLASSNAMES.SELECTED_ITEM);
 
+    // set aria active descendant
+    this.inputField.setAttribute('aria-activedescendant', `item-${index}`);
```
jaspk06, line 395: and if we wanna go with the above suggestion
```suggestion
    this.inputField.setAttribute('aria-activedescendant', `${CLASSNAMES.RESULTS_ITEM}-${index}`);
```

jaspk06 (commented):
is there a tldr on the aria attributes we added?

src/ui/autocomplete.ts
jaspk06: we should prob make this `${CLASSNAMES.RESULTS_ITEM}-${index}` since `id` attributes should be unique. `item-${index}` might cause issues with existing ids

PR #171, opened 2024-06-13
Add core build to CDN output

kochis:


src/core.ts
```diff
@@ -0,0 +1,212 @@
...
+} from './types';
+
+const isSecretKey = (key: string): boolean => (
+  key.includes('_sk_')
+);
+const isLiveKey = (key: string): boolean => (
+  key.includes('_live_')
+);
+
+class Radar {
```
jaspk06, line 44: in `index.ts`, should we do 

```
class Radar extends RadarCore {
  // "ui" namespace
  public static get ui() {
    return {
      maplibregl: MapUI.getMapLibre(),
      map: MapUI.createMap,
      marker: MapUI.createMarker,
      autocomplete: AutocompleteUI.createAutocomplete,
    };
  }
}
```

to minimize code duplication

kochis: Nice

PR #173, opened 2024-07-18
Autocomplete fixes (race condition, prevent autofill, style fixes)

kochis:
* Fixes a race-condition on autocomplete requests
* Disable browser autofill for address autocomplete
* Autocomplete style fixes

## Race Condition
There is a subtle race condition in the autocomplete component that exists when a previous request completes before the latest request. In the scenario where multiple requests are in-flight, the first request to complete sets the results (regardless if the user has typed more characters). The fix is to cancel any in-flight requests when another is issued.

Note: responses are manually throttled down in the examples to demonstrate the issue

#### Race Condition:
https://github.com/user-attachments/assets/69927fc6-688e-4e57-a02d-a80784d0c662

#### Fix:
https://github.com/user-attachments/assets/1821f72a-ea65-4ace-8f07-7e085e857674

### DIFF
--- a/src/ui/autocomplete.ts
+++ b/src/ui/autocomplete.ts
@@ -263,7 +263,7 @@ class AutocompleteUI {
       onRequest(params);
     }
 
-    const { addresses } = await SearchAPI.autocomplete(params);
+    const { addresses } = await SearchAPI.autocomplete(params, 'autocomplete-ui');
     return addresses;
   }
 
--- a/src/ui/autocomplete.ts
+++ b/src/ui/autocomplete.ts
@@ -168,6 +168,9 @@ class AutocompleteUI {
       this.container.appendChild(this.wrapper);
     }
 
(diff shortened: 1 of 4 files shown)


src/ui/autocomplete.ts
```diff
@@ -168,6 +168,9 @@ class AutocompleteUI {
       this.container.appendChild(this.wrapper);
     }
 
+    // disable browser autofill
+    this.inputField.setAttribute('autocomplete', 'off');
+
```
jaspk06, line 173: do we wanna pass this in as an option to in case someone wants to use autofill?

jaspk06 (approved):

src/ui/autocomplete.ts
kochis: I think in that scenario, they can use their own input field.

PR #177, opened 2024-08-01
Add caching for custom markers form URLs

kochis:
* Cache custom markers to avoid unnecessary requests

## before
https://github.com/user-attachments/assets/18cc6ede-46f7-4061-8d4e-68c7063e5eaf

## after

https://github.com/user-attachments/assets/8de9a16a-9cc9-426a-9bc1-28940d554210

### DIFF
--- a/src/ui/RadarMarker.ts
+++ b/src/ui/RadarMarker.ts
@@ -29,6 +29,29 @@ interface ImageOptions {
   height?: number | string;
 }
 
+// cache URL loaded markers
+const IMAGE_CACHE = new Map<string, Blob | string>();
+
+const useCachedImage = (url: string, timeoutMS: number = 5000): Promise<Blob> => new Promise((resolve, reject) => {
+  const start = Date.now();
+  const interval = setInterval(() => {
+    const cachedData = IMAGE_CACHE.get(url);
+    if (cachedData === 'pending') {
+      if (Date.now() - start > timeoutMS) { // expired
+        clearInterval(interval);
+        reject();
+      }
+    } else if (cachedData === 'failed') { // request failed
+      clearInterval(interval);
+      reject();
+
+    } else { // return data
+      clearInterval(interval);
+      resolve(cachedData as Blob);
+    }
+  }, 100);
+});
+
 const createImageElement = (options: ImageOptions) => {
   const element = document.createElement('img');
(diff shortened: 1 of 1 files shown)


src/ui/RadarMarker.ts
```diff
@@ -29,6 +29,29 @@ interface ImageOptions {
   height?: number | string;
 }
 
+// cache URL loaded markers
+const IMAGE_CACHE = new Map<string, Blob | string>();
```
kochis, line 33: TODO: use enum here instead of `string`

src/ui/RadarMarker.ts
```diff
@@ -29,6 +29,29 @@ interface ImageOptions {
...
+
+const useCachedImage = (url: string, timeoutMS: number = 5000): Promise<Blob> => new Promise((resolve, reject) => {
+  const start = Date.now();
+  const interval = setInterval(() => {
+    const cachedData = IMAGE_CACHE.get(url);
+    if (cachedData === 'pending') {
+      if (Date.now() - start > timeoutMS) { // expired
```
kochis, line 40: expired -> timed out

src/ui/RadarMarker.ts
```diff
@@ -97,33 +125,64 @@ class RadarMarker extends maplibregl.Marker {
...
+        // attempt to use cached data, otherwise fetch marker image data from URL
+        useCachedImage(markerOptions.url)
+          .then(onSuccess)
+          .catch((reason: 'miss' | 'timedout' | 'failed' | Error) => {
+            if (reason !== 'miss') {
+              Logger.debug(`RadarMarker: cache lookup for ${markerOptions.url}: ${reason}`);
             }
-          })
-          .catch(onError)
+            loadImage();
```
corypisano, line 158: might be nice to just have `useCachedImage` do the 1) check if in cache 2) if its not, load it

kochis: The only think that makes it sorta tricky is the image loading happens differently depending on the type of marker being loaded (either from a remote URL, or via authenticated request to Radar). Can maybe play around with it.

corypisano: yea maybe could pass in a `loadImage` function, but not a huge deal

PR #178, opened 2024-08-11
Add startTrackingVerified(), getVerifiedLocationToken(), and onTokenUpdated()

nickpatrick:


src/api/verify.ts
```diff
@@ -78,14 +83,86 @@ class VerifyAPI {
       expiresAt,
       expiresIn,
       passed,
+      failureReasons,
+      _id,
```
kochis, line 87: What `_id` is this?

src/api/verify.ts
```diff
@@ -78,14 +83,86 @@ class VerifyAPI {
...
     } as RadarTrackVerifiedResponse;
 
     if (options.debug) {
       trackRes.response = response;
     }
 
+    lastToken = trackRes;
```
kochis, line 94: token is the entire response?

src/api/verify.ts
```diff
@@ -6,7 +6,14 @@ import Logger from '../logger';
 import Session from '../session';
 import Storage from '../storage';
 
-import type { RadarTrackParams, RadarTrackVerifiedResponse } from '../types';
+import type { RadarStartTrackingVerifiedParams, RadarTrackParams, RadarTrackVerifiedResponse } from '../types';
+
+let tokenTimeoutId: any | null = null;
```
kochis, line 11: Should we namespace these as `verified` instead of `token`?

src/api/verify.ts
```diff
@@ -78,14 +83,86 @@ class VerifyAPI {
...
+      failureReasons,
+      _id,
     } as RadarTrackVerifiedResponse;
 
     if (options.debug) {
       trackRes.response = response;
     }
 
+    lastToken = trackRes;
+    lastTokenNow = performance.now();
```
kochis, line 95: We should use `Date.now()` unless there's a need for nano-second measurement (this is really only intended for perf measurement)

src/api/verify.ts
kochis: Also, it's probably preferable to mark things as optional instead of `null`
```ts
let tokenTimeoutId?: any;
```
Otherwise it needs to be explicitly set to `null`

src/api/verify.ts
```diff
@@ -78,14 +83,86 @@ class VerifyAPI {
...
+      }
+
+      // re-request early to maximize the likelihood that a cached token is available
+      if (minInterval > 20) {
+        minInterval = minInterval - 10;
+      }
+
+      // min interval is 10 seconds
+      if (minInterval < 10) {
+        minInterval = 10;
```
david-goodfellow, line 127: so if I give the interval param as 21 then it is effectively 11 but if I give 15 it is effectively 15?

src/api/verify.ts
```diff
@@ -6,7 +6,14 @@ import Logger from '../logger';
...
 
-import type { RadarTrackParams, RadarTrackVerifiedResponse } from '../types';
+import type { RadarStartTrackingVerifiedParams, RadarTrackParams, RadarTrackVerifiedResponse } from '../types';
+
+let tokenTimeoutId: any | null = null;
+let tokenCallback: ((token: RadarTrackVerifiedResponse) => void) | null = null;
+let lastToken: RadarTrackVerifiedResponse | null = null;
+let lastTokenNow: number = 0;
+let expectedCountryCode: string | null = null;
+let expectedStateCode: string | null = null;
```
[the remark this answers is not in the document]
nickpatrick: Don't need this to work across page load like `userId`, so should be good. I could move it inside the `class` if that's important

src/api/verify.ts
nickpatrick: Yeah. I think that's an edge case, but if it bothers us, I can update

src/api/verify.ts
nickpatrick: The problem with `Date.now()` is that you can manipulate it by changing your system clock time, whereas `performance.now()` is based off of time since page load and harder to tamper with

src/api/verify.ts
nickpatrick: `location._id`

src/api/verify.ts
nickpatrick: Yes. Deserialized to `RadarVerifiedLocationToken` in mobile SDKs

src/api/verify.ts
```diff
@@ -78,14 +83,86 @@ class VerifyAPI {
...
+      if (lastTokenElapsed < (lastToken.expiresIn || 0)) {
+        return lastToken;
+      }
+    }
+
+    return this.trackVerified({});
+  }
+
+  static setExpectedJurisdiction(countryCode?: string, stateCode?: string) {
+    expectedCountryCode = countryCode || null;
```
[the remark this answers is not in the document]
nickpatrick: I was getting errors because `expectedCountryCode` is defined as `| null` instead of `| undefined`. https://app.circleci.com/pipelines/github/radarlabs/radar-sdk-js/1241/workflows/a23a44d8-bbed-4f3d-a059-163a7b4407c0/jobs/1016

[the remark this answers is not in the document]
nickpatrick: I was getting errors because `expectedCountryCode` is defined as `| null` instead of `| undefined`. https://app.circleci.com/pipelines/github/radarlabs/radar-sdk-js/1241/workflows/a23a44d8-bbed-4f3d-a059-163a7b4407c0/jobs/1016

src/api/verify.ts
nickpatrick: We do something similar on the mobile SDKs (`System.elapsedRealTimeNanos` on Android) for the same reason

src/api/verify.ts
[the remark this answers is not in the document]
kochis: Yeah, I think you can just define them as:
```ts
let expectedCountryCode?: string;
```

src/api/verify.ts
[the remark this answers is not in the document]
kochis: If these aren't sensitive it's probably fine. Maybe the user-provided ones should be static variables (callback and expected codes)

kochis (approved):

PR #180, opened 2024-09-04
Deprecate countryCode in favor of country param for autocomplete

kochis:
* Deprecates `countryCode` for autocomplete
* Adds `country` param (countryCode will fallback to country for backwards compatibility)


kochis:
Looks like `countryCode` is the preferred param

PR #185, opened 2024-10-10
add skipVerifyApp param to trackVerified() and startTrackingVerified()

nickpatrick:


david-goodfellow (approved):
LGTM!

src/api/verify.ts
```diff
@@ -16,13 +17,15 @@ let expectedCountryCode: string | null = null;
...
-  static async trackVerified(params: RadarTrackParams, encrypted: Boolean = false) {
+  static async trackVerified(params: RadarTrackVerifiedParams, encrypted: Boolean = false) {
     const options = Config.get();
 
+    const { skipVerifyApp } = params;
+
     // user indentification fields
     const userId = params.userId || Storage.getItem(Storage.USER_ID);
-    const deviceId = params.deviceId || Device.getDeviceId();
-    const installId = params.installId || Device.getInstallId();
```
kochis, line 25: Why the change to get rid of setting these via params? This might be how we override the values int the simulator.

src/api/verify.ts
```diff
@@ -103,7 +116,9 @@ class VerifyAPI {
 
   static async startTrackingVerified(params: RadarStartTrackingVerifiedParams) {
     const doTrackVerified = async () => {
-      const trackRes = await this.trackVerified({});
+      const trackRes = await this.trackVerified({
+        ...params,
```
kochis, line 120: nit: can probably just pass `params` instead of destructuring

nickpatrick: Good call, updated

src/api/verify.ts
nickpatrick: I don't really want to give developers control of these for fraud / geo-compliance use cases. For implementations using the Mac and Windows app, those apps override these anyway (with more stable hardware device IDs)

kochis (dismissed):

PR #186, opened 2024-10-16
Add spell check to precommit hook

jaspk06:
- also fixes typos in docs


cspell.json
jaspk06, line 1: taken from https://github.com/radarlabs/documentation

PR #188, opened 2024-11-04
prevent /config call on multiple initializations

kochis:
* Warn on multiple initializations and prevent multiple calls to `/config`

## Before
image

## After
image

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -58,6 +58,10 @@ class Radar {
       throw new RadarPublishableKeyError('Secret keys are not allowed. Please use your Radar publishable key.');
     }
 
+    if (Config.isInitialized()) {
+      Logger.error('Radar.initialize() called more than once.');
+    }
+
     // store settings in global config
     const live = isLiveKey(publishableKey);
     const logLevel = live ? 'error' : 'info';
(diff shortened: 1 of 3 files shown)


src/api.ts
```diff
@@ -58,6 +58,10 @@ class Radar {
       throw new RadarPublishableKeyError('Secret keys are not allowed. Please use your Radar publishable key.');
     }
 
+    if (Config.isInitialized()) {
+      Logger.error('Radar.initialize() called more than once.');
```
lmeier, line 62: Given that we're handling this, should we `Logger.warn` instead?

src/api.ts
```diff
@@ -79,11 +83,15 @@ class Radar {
       Logger.debug('using options', options);
     }
 
-    // NOTE(jasonl): this allows us to run jest tests
-    // without having to mock the ConfigAPI.getConfig call
-    if (!(window as any)?.RADAR_TEST_ENV) {
-      ConfigAPI.getConfig();
+    if (!Config.isInitialized()) { // only call getConfig on first initialization
```
lmeier, line 86: Maybe we want to `|| keyChanged` to handle cases where someone would change the key (can't come up with anything off the top of my head where that would matter here, but we do allow the key to change in Waypoint for instance)

kochis: Yeah, that's the one scenario I was sort of thinking of. 

Maybe it's worth just checking if the config options have changed or not.

src/api.ts
kochis: I agree it's a "warn" and not an error, but I also sort of want this to fire on a production site. Maybe I will just change to `console.warn` instead.

PR #190, opened 2024-11-18
save user._id to storage after track, and use in conversion event

ShiCheng-Lu:


src/api/config.ts
```diff
@@ -16,12 +17,14 @@ class ConfigAPI {
...
 
+    const id = Storage.getItem(Storage.ID) || undefined;
     const deviceId = params.deviceId || Device.getDeviceId();
     const installId = params.installId || Device.getInstallId();
     const sessionId = Session.getSessionId();
     const locationAuthorization = await Navigator.getPermissionStatus();
 
-    const data = {
+    const data: any = {
+      id,
```
kochis, line 27: Just to verify, `id` is the correct param (not `_id`)?

PR #193, opened 2024-12-15
add IP change listener to getVerifiedLocationToken()

nickpatrick:


src/api/track.ts
```diff
@@ -109,31 +108,9 @@ class TrackAPI {
         },
       });
 
-      let sclVal = -1;
```
nickpatrick, line 112: Remove old ping call

src/api/verify.ts
```diff
@@ -15,8 +15,25 @@ let lastToken: RadarTrackVerifiedResponse | null = null;
 let lastTokenNow: number = 0;
 let expectedCountryCode: string | null = null;
 let expectedStateCode: string | null = null;
+let lastIp: string | null = null;
 
 class VerifyAPI {
+  static async ipChanged() {
```
nickpatrick, line 21: Calls `GET /v1/ping`, returns `true` if IP changed

src/api/verify.ts
```diff
@@ -115,22 +132,31 @@ class VerifyAPI {
   }
 
   static async startTrackingVerified(params: RadarStartTrackingVerifiedParams) {
+    let { interval } = params;
+    if (!interval) {
+      interval = 60;
+      Logger.info(`interval not provided, using 60 seconds`);
```
nickpatrick, line 138: Adds default interval of 60 seconds, avoids bug if interval undefined

src/api/verify.ts
```diff
@@ -164,7 +190,12 @@ class VerifyAPI {
   static async getVerifiedLocationToken(params: RadarTrackVerifiedParams) {
     const lastTokenElapsed = (performance.now() - lastTokenNow) / 1000;
 
-    if (lastToken && lastToken.passed) {
+    let ipChanged = false;
+    if (params.ipChanges) {
+      ipChanged = await this.ipChanged();
+    }
+
+    if (lastToken && lastToken.passed && !ipChanged) {
```
nickpatrick, line 198: Requests new token if IP changed

src/ui/RadarMap.ts
```diff
@@ -96,7 +96,7 @@ class RadarMap extends maplibregl.Map {
       radarMapOptions,
       { style },
     );
-    Logger.debug('map initailized with options', mapOptions);
+    Logger.debug('map initialized with options', mapOptions);
```
nickpatrick, line 99: Fix typo

src/api/verify.ts
```diff
@@ -104,6 +121,11 @@ class VerifyAPI {
       }
     }
 
+    if (params.ipChanges && trackRes?.user?.ip) {
```
nickpatrick, line 124: Also picks up IP changes in trackVerified() responses

src/api/verify.ts
kochis: nit: would call this function `checkIPChange`

Also, should this function only fire if `lastIp` is set? otherwise it will always trigger a change

PR #196, opened 2025-01-22
avoid double request, add clearVerifiedLocationToken()

nickpatrick:
Android equivalent: https://github.com/radarlabs/radar-sdk-android/pull/426
iOS equivalent: https://github.com/radarlabs/radar-sdk-ios/pull/427

### DIFF
--- a/src/api/verify.ts
+++ b/src/api/verify.ts
@@ -114,31 +114,23 @@ class VerifyAPI {
     return trackRes;
   }
 
-  static async startTrackingVerified(params: RadarStartTrackingVerifiedParams) {
-    const doTrackVerified = async () => {
-      let trackRes;
-      try {
-        trackRes = await this.trackVerified(params);
-      } catch (err: any) {
-        Logger.error(`trackVerified error: ${err.message}`);
-      }
-      
+  static startTrackingVerified(params: RadarStartTrackingVerifiedParams) {
+    const scheduleNextIntervalWithLastToken = async () => {
       const { interval } = params;
 
-      let expiresIn = 0;
       let minInterval = interval;
 
-      if (trackRes) {
-        expiresIn = (trackRes.expiresIn || expiresIn);
+      if (lastToken) {
+        const lastTokenElapsed = (performance.now() - lastTokenNow) / 1000;
+
+        const expiresIn = (lastToken.expiresIn || 0);
 
         // if expiresIn is shorter than interval, override interval
-        minInterval = Math.min(expiresIn, interval);
+        // re-request early to maximize the likelihood that a cached token is available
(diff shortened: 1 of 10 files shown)


src/api/verify.ts
```diff
@@ -114,31 +114,23 @@ class VerifyAPI {
...
+  static startTrackingVerified(params: RadarStartTrackingVerifiedParams) {
+    const scheduleNextIntervalWithLastToken = async () => {
       const { interval } = params;
 
-      let expiresIn = 0;
       let minInterval = interval;
 
-      if (trackRes) {
-        expiresIn = (trackRes.expiresIn || expiresIn);
+      if (lastToken) {
```
kochis, line 123: Where is `lastToken` in scope here?

kochis: Ah, I see it's defined at the top of the file.

nickpatrick: That's right. It stores the last `trackVerified()` response (if any)

src/api/verify.ts
```diff
@@ -114,31 +114,23 @@ class VerifyAPI {
...
-        minInterval = Math.min(expiresIn, interval);
+        // re-request early to maximize the likelihood that a cached token is available
+        minInterval = Math.min(expiresIn - lastTokenElapsed, interval);
       }
 
-      // re-request early to maximize the likelihood that a cached token is available
-      if (minInterval > 20) {
-        minInterval = minInterval - 10;
-      }
+      minInterval = minInterval - 10;
```
kochis, line 133: The one thing that is hard to follow in this code is what things are seconds, and what are milliseconds. I wonder if we should rename this to:
```
minIntervalSeconds
```
Just to make it more clear (as well as other things that might be in seconds / ms)

kochis: Also, is there a reason we refer to it as `lastToken` instead of `lastResponse`?

src/api/verify.ts
kochis: Is there a reason we refer to it as `lastToken` instead of lastResponse?

kochis (approved):

src/api/verify.ts
nickpatrick: It's fair feedback. In this case I was just trying to mirror the mobile SDK code + var names as closely as possible to make it easy to compare across platforms

src/api/verify.ts
nickpatrick: Just mirroring var names in the mobile SDK to make it easy to compare across platforms. On the SDK, the response `{ passed, token, expiresIn, ... }` is deserialized to `RadarVerifiedLocationToken`

PR #203, opened 2025-04-15
Autocomplete fallback font-family

jaspk06:
- Adds a sans-serif font fallback

image

### DIFF
--- a/styles/radar.css
+++ b/styles/radar.css
@@ -36,7 +36,7 @@
 
 .radar-autocomplete-wrapper {
   position: relative;
-  font-family: 'Graphik';
+  font-family: 'Graphik', sans-serif;
 }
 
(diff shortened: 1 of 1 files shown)


boonboonsiri (approved):

boonboonsiri:
This is a good test for whether static.radar.com also has any CORS headers

jaspk06:
> This is a good test for whether static.radar.com also has any CORS headers (which looking at it I don't think it has and I think it needs)

@boonboonsiri yeah looks like it does

boonboonsiri:
Looks like it has CORS header or looks like it needs?

jaspk06:
> Looks like it has CORS header or looks like it needs?

@kochis figured it out

kochis (approved):

PR #205, opened 2025-05-09
add fraud product support

JoRoRadar:
Add get/set methods for product type and add product header to verify calls.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -237,6 +237,18 @@ class Radar {
   public static matrix(params: RadarMatrixParams) {
     return RoutingAPI.matrix(params);
   }
+
+  ///////////////////////
+  // Fraud
+  ///////////////////////
+
+  public static setFraudProduct(product?: string) {
+    if (!product) {
+      Storage.removeItem(Storage.FRAUD_PRODUCT);
+      return;
+    }
+    Storage.setItem(Storage.FRAUD_PRODUCT, String(product).trim());
+  }
 }
 
 export default Radar;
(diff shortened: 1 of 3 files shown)


src/api.ts
```diff
@@ -237,6 +237,18 @@ class Radar {
   public static matrix(params: RadarMatrixParams) {
     return RoutingAPI.matrix(params);
   }
+
+  ///////////////////////
+  // Fraud
+  ///////////////////////
+
+  public static setFraudProduct(product?: string) {
```
kochis, line 245: I think we just call this `setProduct` in the mobile SDKs

kochis:
Since this isn't really fraud-specific, lets remove the fraud namespace everywhere

src/api/track.ts
```diff
@@ -66,6 +66,13 @@ class TrackAPI {
       tripOptions.version = '2';
     }
 
+    var reqHeaders: { 'X-Radar-Product'?:string } = {}
```
kochis, line 69: Lets define the type at the top of this file with the other types:
```ts
interface TrackRequestHeaders {
  'X-Radar-Body-Is-Token'?: boolean;
  'X-Radar-Product'?: string;
}
```

Then we can create the object like:
```ts
const headers: TrackRequestHeaders = {};

if (fraud) {
  headers['X-Radar-Body-Is-Token'] = true;
}

if (product) {
  headers['X-Radar-Product'] = product;
}
```

src/api/track.ts
```diff
@@ -66,6 +71,13 @@ class TrackAPI {
       tripOptions.version = '2';
     }
 
+    var headers: TrackRequestHeaders = {}
```
kochis, line 74: nit: `const`

kochis (approved):

PR #206, opened 2025-05-13
FRA-1076: add optional reason to track params

mai-radar:
Adding an optional `reason` param to track params

Use Case:
```
Radar.track({ userId: 'test-user-id', reason: 'test-reason'})
```

Including reasons in the trackOnce request:
Screenshot 2025-05-13 at 1 04 20 PM

Not including reasons:
Screenshot 2025-05-13 at 1 53 36 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "radar-sdk-js",
-      "version": "4.5.2",
+      "version": "4.5.0",
       "license": "ISC",
       "dependencies": {
         "@types/geojson": "^7946.0.10"
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "radar-sdk-js",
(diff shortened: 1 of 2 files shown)


mai-radar:
actually it looks like you can add a reason parameter currently without any changes due to structural typing in Typescript, but that brings up the fact that the types defined in `type.ts` allows for anyone to add any param to the request, which I'm not sure we want?

kochis:
> actually it looks like you can add a reason parameter currently without any changes due to structural typing in Typescript, but that brings up the fact that the types defined in type.ts allows for anyone to add any param to the request, which I'm not sure we want?

Yeah, typescript doesn't actually enforce anything - you can technically use this library without typescript. It's really just there to ensure type cohesion is correct when used in a typescript codebase (also provides hints to parameters and their expected values).  All the parameter validation is done server-side.

package-lock.json
kochis, line 1: There probably shouldn't be any `package-lock` changes here

kochis (approved):

PR #209, opened 2025-07-01
swallow error and return locationAuthorization = NOT_DETERMINED when navigator.permissions is not supported

nickpatrick:


src/api/config.ts
```diff
@@ -19,7 +19,14 @@ class ConfigAPI {
     const deviceId = params.deviceId || Device.getDeviceId();
     const installId = params.installId || Device.getInstallId();
     const sessionId = Session.getSessionId();
-    const locationAuthorization = await Navigator.getPermissionStatus();
+
+    // location authorization
+    let locationAuthorization;
```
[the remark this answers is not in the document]
nickpatrick: @kochis @jaspk06 Looks like `locationAuthorization` is optional in `GET /config`, so it would be fine if this is `undefined` in rare cases. `NOT_DETERMINED` is the default inside of `getPermissionsStatus()`

src/api/config.ts
```diff
@@ -19,7 +19,14 @@ class ConfigAPI {
...
+    try {
+      locationAuthorization = await Navigator.getPermissionStatus();
+    } catch (err: any) {
+      Logger.warn(`Location authorization error: ${err.message}`);
+    }
```
jaspk06, line 29: nit: i would refactor this a bit to make it less verbose esp since `getPermissionStatus` _should_ technically no longer throw

```suggestion
      const locationAuthorization = await Navigator.getPermissionStatus().catch((err: any) => {
      Logger.warn(`Location authorization error: ${err.message}`);
      }
```

but im happy to take this across the line

jaspk06 (approved):
cool can def get on board with this esp since we're just sending up `NOT_DETERMINED` which aligns with the rest of our SDKs right. Happy to take the nit and release across the line

src/api/config.ts
kochis: I would leave as `try/catch`. Chaining the catch is generally for fire-and-forget requests. We're still using `await` here.

jaspk06: sure just a code styling thing. ik we use it in a bunch of other async code paths with `await` since it removes the need for another level of nesting but not super opinionated here. just makes things a bit cleaner

PR #215, opened 2025-09-17
Update RadarAddress type 

alanjcharles14:
- Adds optional `categories` array to `RadarAddress`


lmeier:
@alanjcharles14 do we only need the type bump? Seeing some QA for this would be helpful too 😄

lmeier (dismissed):

PR #257, opened 2026-06-17
fix(http): proper RFC 3986 encoding in GET requests

jaspk06:
## Problem
reason = "Game Launch" is picked up by Verify apps as "Game+Launch"

we were implicitly encoding our GET query params through `application/x-www-form-urlencoded` rather than `RFC 3986`

when you do something like:
```
> params = new URLSearchParams({ reason: 'game launch' })
URLSearchParams { 'reason' => 'game launch' }
> params.toString()
'reason=game+launch'
```
we'll have to replace `+` with `%20`. this is the main deviation between the two encoding protocols

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -112,7 +112,7 @@ class Http {
       const params = new URLSearchParams(Object.entries(filtered).map(([k, v]) => [k, String(v)]));
       // URLSearchParams serializes spaces as '+' (application/x-www-form-urlencoded)
       // a query string is part of a URI, so encode per RFC 3986 (%20)
-      // the only differnce is really how spaces are handled between the two protocols
+      // the only difference is really how spaces are handled between the two protocols
       const qs = params.toString().replace(/\+/g, '%20');
       if (qs) {
         url = `${url}?${qs}`;
--- a/src/http.ts
+++ b/src/http.ts
@@ -110,7 +110,10 @@ class Http {
 
     if (method === 'GET') {
       const params = new URLSearchParams(Object.entries(filtered).map(([k, v]) => [k, String(v)]));
-      const qs = params.toString();
+      // URLSearchParams serializes spaces as '+' (application/x-www-form-urlencoded)
+      // a query string is part of a URI, so encode per RFC 3986 (%20)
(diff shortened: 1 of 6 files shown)


test/http.test.ts
```diff
@@ -131,11 +131,95 @@ describe('Http', () => {
       const response = await Http.request({ method: 'GET', path: 'geocode/forward', data });
       const request = getRequest();
 
-      expect(request.url).toContain('?query=841+Broadway');
+      expect(request.url).toContain('?query=841%20Broadway');
```
jaspk06, line 134: ya i guess this was an oversight 🙃

test/http.test.ts
```diff
@@ -131,11 +131,95 @@ describe('Http', () => {
...
+    });
+
+    it('should preserve a literal + as %2B', async () => {
+      mockRequest(200, successResponse);
+
+      await Http.request({ method: 'GET', path: 'geocode/forward', data: { query: '1+1 = 2' } });
+      const request = getRequest();
+
+      expect(request.url).toContain('?query=1%2B1%20%3D%202');
+    });
```
jaspk06, line 156: the tests that we care most ab - the other tests were added for better RFC 3986 coverage

src/http.ts
```diff
@@ -110,7 +110,10 @@ class Http {
 
     if (method === 'GET') {
       const params = new URLSearchParams(Object.entries(filtered).map(([k, v]) => [k, String(v)]));
-      const qs = params.toString();
+      // URLSearchParams serializes spaces as '+' (application/x-www-form-urlencoded)
+      // a query string is part of a URI, so encode per RFC 3986 (%20)
+      // the only differnce is really how spaces are handled between the two protocols
```
exogen:
```suggestion
      // the only difference is really how spaces are handled between the two protocols
```

exogen (dismissed):

src/http.ts
jaspk06: co author 🤝

exogen (dismissed):

Luke-zhang-04 (dismissed):

exogen (approved):

PR #259, opened 2026-07-07
FRA-2081 revealRisk path - DRAFT

siebert-brian:
This will require some additional clean-up and unit testing.

### DIFF
--- a/src/api/reveal-risk.ts
+++ b/src/api/reveal-risk.ts
@@ -1,12 +1,12 @@
-import Config from "../config";
-import Device from "../device";
-import Http from "../http";
-import Logger from "../logger";
-import Session from "../session";
-import Storage from "../storage";
-import SDK_VERSION from "../version";
+import Config from '../config';
+import Device from '../device';
+import Http from '../http';
+import Logger from '../logger';
+import Session from '../session';
+import Storage from '../storage';
+import SDK_VERSION from '../version';
 
-import type { RadarRevealRiskParams, RadarRevealRiskResponse } from "../types";
+import type { RadarRevealRiskParams, RadarRevealRiskResponse } from '../types';
 
 class RevealRiskAPI {
   /**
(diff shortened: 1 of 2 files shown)


jaspk06:
@siebert-brian i think this should live in the fraud sdk but lemme double check

src/api/reveal-risk.ts
```diff
@@ -0,0 +1,78 @@
...
+    const sessionId = Session.getSessionId();
+    const deviceType = params.deviceType || 'Web';
+    const description = params.description || Storage.getItem(Storage.DESCRIPTION);
+
+    // save userId for trip tracking
+    if (!userId) {
+      Logger.warn('userId not provided for trackOnce.');
+    } else {
+      Storage.setItem(Storage.USER_ID, userId);
+    }
```
jaspk06, line 33: dont think this is relevant for reveal risk

jaspk06: well i guess more so the first logger warn - we don't realllllyyy care that much ab userId here

jaspk06: but yeah i guess it's like if userId exists then set it in storage

src/api/reveal-risk.ts
```diff
@@ -0,0 +1,78 @@
...
+      foreground: true,
+      installId,
+      sessionId,
+      metadata,
+      sdkVersion: SDK_VERSION,
+      stopped: true,
+      userId,
+    };
+
+    const response = await Http.request<Omit<RadarRevealRiskResponse, 'response' | 'reveal/risk'>>({
```
jaspk06, line 52: whats this second thing we're omitting

src/api/reveal-risk.ts
siebert-brian: ok removing

src/api/reveal-risk.ts
siebert-brian: I based the request off what was in track:
```
    const response = await Http.request<Omit<RadarTrackResponse, 'response' | 'location'>>({
      method: 'POST',
      path: 'track',
      data: body,
    });
```
what do you think can be improved here?

jaspk06: well is that second field we're omitting even in the response?

PR #262, opened 2026-08-11
SDK Timeouts

boonboonsiri:
image

configurable timeout

### DIFF
--- a/src/http.ts
+++ b/src/http.ts
@@ -69,6 +69,42 @@ const getNetworkTimeoutInterval = (interval?: number) => {
   return Math.min(Math.max(timeout, MIN_NETWORK_TIMEOUT_INTERVAL), MAX_NETWORK_TIMEOUT_INTERVAL);
 };
 
+type AbortSignalConstructor = typeof AbortSignal & {
+  any?: (signals: AbortSignal[]) => AbortSignal;
+  timeout?: (milliseconds: number) => AbortSignal;
+};
+
+const getAbortSignalConstructor = (): AbortSignalConstructor | undefined => {
+  if (typeof AbortSignal === 'undefined') {
+    return undefined;
+  }
+
+  return AbortSignal as AbortSignalConstructor;
+};
+
+const createTimeoutSignal = (timeoutMS: number): AbortSignal | undefined => {
+  const AbortSignalCtor = getAbortSignalConstructor();
+  return AbortSignalCtor?.timeout?.(timeoutMS);
+};
+
+const combineAbortSignals = (signals: AbortSignal[]): AbortSignal | undefined => {
+  if (signals.length === 0) {
+    return undefined;
+  }
+
+  if (signals.length === 1) {
+    return signals[0]!;
+  }
+
+  const AbortSignalCtor = getAbortSignalConstructor();
+  if (AbortSignalCtor?.any) {
+    return AbortSignalCtor.any(signals);
+  }
+
+  // Without AbortSignal.any(), prefer requestId cancellation over timeout when both signals exist.
+  return signals[0]!;
+};
+
 /** fetch-based HTTP client for Radar API requests */
 class Http {
   /** map of host patterns to custom error factories for intercepting network errors */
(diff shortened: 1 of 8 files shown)


src/http.ts
```diff
@@ -55,6 +55,20 @@ interface HttpRequestOptions {
...
+const MAX_NETWORK_TIMEOUT_INTERVAL = 300;
+
+const getNetworkTimeoutInterval = (interval?: number) => {
+  let timeout = interval ?? DEFAULT_NETWORK_TIMEOUT_INTERVAL;
+
+  if (timeout <= 0 || !Number.isFinite(timeout)) {
+    timeout = DEFAULT_NETWORK_TIMEOUT_INTERVAL;
+  }
+
+  return Math.min(Math.max(timeout, MIN_NETWORK_TIMEOUT_INTERVAL), MAX_NETWORK_TIMEOUT_INTERVAL);
```
boonboonsiri, line 76: https://github.com/radarlabs/radar-sdk-ios/blob/af76ca52f32e2c5bad6640bf9bbf2e536dd52388/RadarSDK/RadarAPIHelper.m#L16-L66

src/http.ts
```diff
@@ -127,6 +141,12 @@ class Http {
     }
 
     const abortController = new AbortController();
+    const timeoutId = setTimeout(
+      () => {
+        abortController.abort();
+      },
+      getNetworkTimeoutInterval(options.networkTimeoutInterval) * 1000,
+    );
```
jaspk06, line 193: curious if this is cleaner if we use an AbortSignal instead so that we dont have to race an internal timeout thing

Image

jaspk06: i also wonder if this means we can kill our "in flight requests" map thing thats a bit sketchy too

src/version.ts
```diff
@@ -1 +1 @@
-export default '5.1.1';
```
jaspk06: i guess this is a feat? or a fix? i wonder if 10s makes this a breaking change or not (prob not)

src/http.ts
boonboonsiri: Sure I'm good with abort signal I mean I will always prefer the "less cleanup" code option as I think it looks cleaner. I Think we still ned the inflight request thing for autocomplete UI unfortunately

src/version.ts
boonboonsiri: I saw this as a feat. uh I can't imagine what request would be taking 10s I swear a lot of our timeouts in server (especially for mongo) are like 5s

boonboonsiri:
Reverified that this works

image

src/http.ts
jaspk06: ah yeah forgot ab the autocomplete ui quirk

src/version.ts
jaspk06: yeah well if a bunch of 5s timeouts compound then we can easily surpass 10s

jaspk06: but yeah thatll only happen if something real bad goes wrong

src/http.ts
```diff
@@ -62,6 +62,56 @@ interface HttpRequestOptions {
...
+    timeout = DEFAULT_NETWORK_TIMEOUT_INTERVAL;
+  }
+
+  return Math.min(Math.max(timeout, MIN_NETWORK_TIMEOUT_INTERVAL), MAX_NETWORK_TIMEOUT_INTERVAL);
+};
+
+type AbortSignalConstructor = typeof AbortSignal & {
+  any?: (signals: AbortSignal[]) => AbortSignal;
+  timeout?: (milliseconds: number) => AbortSignal;
+};
```
jaspk06, line 83: worth adding a note that we're leaving the optional cuz of browser compatibility

jaspk06 (dismissed):

jaspk06:
and prob worth just updating the PR name to just have it scoped to timeouts cuz technically it doesnt have to do with retries

boonboonsiri:
Okay took a look at timeouts to make sure they are good in case. I realize ya because this is a feature should technically be a 5.2.0 not super happy about this feels like a "big" version bump for such a small thing.

Also makes me question our sdk versioning a bit. I feel like there's no way we only released like 5 features on our js sdk in like 2 years

image


image


Couldn't find an easy way to test with an old browser though

jaspk06 (approved):

PR #263, opened 2026-08-20
MAPS-2581: Add autocomplete session support

binhrobles:
- Adds AutocompleteUI support for autocomplete session creation and clickthrough events
- Adds `requestId` to `RadarApiMeta` object, for easier CS debugging flows

### QA
/autocomplete calls include `sessionToken`:
Screenshot 2026-08-24 at 1 19 44 PM

click event sends /click request:
Screenshot 2026-08-24 at 1 19 35 PM

20th /click event returns 429, and debug logs, but does not disrupt client functionality:
Screenshot 2026-08-24 at 1 19 04 PM

### DIFF
--- a/src/api/search.ts
+++ b/src/api/search.ts
@@ -34,9 +34,7 @@ class SearchAPI {
       }
     }
 
-    const { data: response, requestId: radarRequestId } = await Http.request<
-      Omit<RadarAutocompleteResponse, 'response' | 'requestId'>
-    >({
+    const response = await Http.request<Omit<RadarAutocompleteResponse, 'response' | 'requestId'>>({
       method: 'GET',
       path: 'search/autocomplete',
       data: {
@@ -52,13 +50,12 @@ class SearchAPI {
         sessionToken,
       },
       requestId,
-      includeRequestId: true,
     });
 
     const autocompleteRes: RadarAutocompleteResponse = {
       addresses: response.addresses,
       // echoed back to search/autocomplete/click to attribute a selection to this response
-      requestId: radarRequestId,
+      requestId: response.meta?.requestId,
     };
 
     if (options.debug) {
--- a/src/api/search.ts
+++ b/src/api/search.ts
@@ -76,7 +76,7 @@ class SearchAPI {
         method: 'POST',
         path: 'search/autocomplete/click',
         data: { sessionToken, requestId, idx },
-        // the selection often navigates the page; keepalive lets the report outlive it
+        // the selection may navigate the page; keepalive lets the report outlive it
         keepalive: true,
       });
     } catch (err) {
--- a/src/api/search.ts
+++ b/src/api/search.ts
@@ -1,5 +1,6 @@
 import Config from '../config';
 import Http from '../http';
+import Logger from '../logger';
 import Navigator from '../navigator';
 
 import type {
@@ -78,9 +79,10 @@ class SearchAPI {
         // the selection often navigates the page; keepalive lets the report outlive it
         keepalive: true,
       });
-    } catch {
+    } catch (err) {
       // clickthroughs are fire-and-forget. an expired session or a network blip must never
-      // surface to the caller, and is never retried.
+      // surface to the caller, and is never retried
+      Logger.debug(`Autocomplete click not reported: ${String(err)}`);
     }
   }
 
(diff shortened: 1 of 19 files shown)


src/http.ts
```diff
@@ -24,6 +24,11 @@ interface RadarApiMeta {
   error?: string;
   message?: string;
   type?: string;
+  /**
+   * Attached from the `x-radar-request-id` response header
+   * For use when reporting an issue to Radar
+   */
+  requestId?: string;
```
binhrobles, line 31: Maybe controversial, but I think this is the best of a few meh options.

This makes our client SDK `RadarApiMeta` different than our actual response JSON's `meta`, because we're attaching the requestId here. 

I think this is better than:
- attaching it anywhere in the RadarApiResponse type
  - because it really doesn't exist there
- modifying our existing Http.request fn signature to include a `requestId`
  - because that's a breaking change at the fn's top level
- adding an Http.request overload with a `includeRequestId` param + response
  - because that's an ugly API
- adding `requestId` in the actual response `meta` object
  - because it's already in the `x-radar-request-id` header, and we'd need to add particulars around endpoints that cache

Adding it to the `RadarApiMeta` means that consumers "just get it" and, if they're logging their responses, will be able to reference them for support tickets immediately, rather than needing to change code.

Will want to add this to our other SDKs as well

Think this fits in with how other SDKs use a metadata object on their response, and even though this isn't purely coming from the response body anymore, it's still consistent with how we communicate this sort of info.

tsconfig.test.json
```diff
@@ -3,7 +3,14 @@
   "compilerOptions": {
     "moduleResolution": "node",
     "noUnusedLocals": false,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "baseUrl": ".",
+    // Same remap as jest's moduleNameMapper, but for the typechecker rather than the
```
mohamed-rdr, line 8: json doesn't allow comments?

src/api/search.ts
```diff
@@ -61,6 +65,25 @@ class SearchAPI {
...
+    try {
+      await Http.request({
+        method: 'POST',
+        path: 'search/autocomplete/click',
+        data: { sessionToken, requestId, idx },
+        // the selection often navigates the page; keepalive lets the report outlive it
+        keepalive: true,
+      });
+    } catch {
+      // clickthroughs are fire-and-forget. an expired session or a network blip must never
```
mohamed-rdr, line 83: should we do some logging here?

tsconfig.test.json
binhrobles: `tsconfig*.json` files are actually parsed as JSONC!
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-1-8.html#allow-comments-in-tsconfigjson

src/http.ts
mohamed-rdr: Should we do something like
```ts
type RadarResponseMetadata = {
  requestId?: RadarRequestId;
};

type HttpResult<T> = {
  data: T;
  metadata: RadarResponseMetadata;
};
```

and add `Http.requestWithMetadata()` while retaining `Http.request()` for compatibility.

src/api/search.ts
binhrobles: added

src/http.ts
binhrobles: I'm not sure, because our Radar HTTP responses themselves have the `meta` field. 

so, if I'm understanding correctly, this would be adding a wrapper around our existing fn (similar to the overload option 3 in my prev comment) while producing a data shape like:
```
result { 
  metadata: {
    requestId
  },
  data: {
    meta: {
      statusCode: 200
    },
    ...body...
  }
}
```

which I think is a bit confusing.

Honestly, the fact that we already return a `meta` field and it includes redundant information like `statusCode` makes me want to just return the requestId in there. And invert the middleware implementation so that instead of adding requestId to all endpoints and then removing it for those endpoints that need ETag caching, we only attach it to those endpoints that we don't need ETag caching.

mohamed-rdr: Yeah, that's cool, I can see it both ways

mohamed-rdr (dismissed):

alanjcharles14 (dismissed):

alanjcharles14 (approved):

